PR #27, opened 2020-02-17
Add styles for dashboard

richtier:
 - [x] Changelog entry added.
 
image


core/sass/base.scss
```diff
@@ -1,63 +1,123 @@
 @import 'helpers/_colours';
 
+.m-l-0 {
```
eadpearce: we have this already it's `m-f-0`

eadpearce: (reasoning was `l` was already used as abbreviation for `large` but since it's in a different position in the class name we could just have them as the same letter)

PR #30, opened 2020-02-18
Mvp 102 select commodity code 2

depsiatwal:
 - [x] Changelog entry added.
 - [x] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [x] (if adding env vars) Added any new environment variable to vault.
 - [x] (if adding feature flags) Cleaned up old flags


core/forms.py
```diff
@@ -0,0 +1,10 @@
+from django import forms
+
+from core import helpers
+
+
+class ExportPlanFormStart(forms.Form):
+    COUNTRY_CHOICE = helpers.get_madb_country_list()
```
richtier: `COUNTRY_CHOICES` is more accurate

it would be simpler if the `*_CHOICES` were moved to module level rather than nested in the form class. _Flat is better than nested_

core/helpers.py
```diff
@@ -34,3 +35,53 @@ def store_user_location(request):
     )
     if not response.ok:
         logger.error(USER_LOCATION_CREATE_ERROR)
+
+
+def get_madb_country_list():
+    airtable = Airtable('appcxR2dZGyugfvyd', 'CountryDBforGIN')
```
richtier: oops. secrets committed to repo :(

please do the needful and reset the credentials.

core/helpers.py
```diff
@@ -34,3 +35,53 @@ def store_user_location(request):
...
+    airtable_data = airtable.get_all(view='Grid view')
+    country_list = [c['Country'] for c in [f['fields'] for f in airtable_data]]
+    return list(zip(country_list, country_list))
+
+
+def get_madb_commodity_list():
+    airtable = Airtable('appcxR2dZGyugfvyd', 'CountryDBforGIN')
+    commodity_name_set = set()
+    for d in airtable.get_all(view='Grid view'):
```
richtier: `d` is bit hard to read. something like `row` would be easier.

rule of thumb: single character var names are ok when doing single line things like list comprehension, otherwise more verbose variable names aids readability

core/helpers.py
```diff
@@ -34,3 +35,53 @@ def store_user_location(request):
...
+        Commodity_code = d['fields']['Commodity code']
```
richtier: snake case please, not stating with uppercase

core/helpers.py
```diff
@@ -34,3 +35,53 @@ def store_user_location(request):
...
+    return commodity_name_set
+
+
+def get_rules_and_regulations(country):
+    airtable = Airtable('appcxR2dZGyugfvyd', 'CountryDBforGIN')
+    rules = airtable.search('country', country)
+    if rules:
+        return rules[0]['fields']
+    else:
+        return None
```
richtier:
```
    else:
        return None
```
is redundant as `None` would be returned by default

core/helpers.py
```diff
@@ -34,3 +35,53 @@ def store_user_location(request):
...
+
+
+def create_export_plan(sso_session_id, exportplan_data):
+    data = serialize_exportplan_data(exportplan_data)
```
richtier: I think `data = serialize_exportplan_data(exportplan_data)` should happen outside of this function, as currently the `create_export_plan` is also transforming form data. it should not really know about he structure of the form.

I think it would be better to move it to the view that calls `create_export_plan` as the view needs to know about the structure of the form ayway

core/helpers.py
```diff
@@ -34,3 +35,53 @@ def store_user_location(request):
...
+def serialize_exportplan_data(exportplan_data):
+    return {
+        'export_countries': [exportplan_data['Country']],
+        'export_commodity_codes': [exportplan_data['Commodity code']],
+        'rules_regulations': exportplan_data,
+    }
+
+
+def get_exportplan_rules_regulations(sso_session_id):
+    exportplan_list = api_client.exportplan.exportplan_list(sso_session_id)
```
richtier: `exportplan_list` is a `response` really. there is no guarantee the response is 200, so would be worth doing `response.raise_for_status()`

core/templates/ExportPlanStart.html
```diff
@@ -0,0 +1,49 @@
+{% load static from staticfiles %}
```
richtier: solid UI :)

core/views.py
```diff
@@ -1,4 +1,41 @@
...
+    success_url = reverse_lazy('core:exportplan-start')
+
+    def get_initial(self):
+        return {
+            'country': self.request.GET.get('country'),
+            'commodity': self.request.GET.get('commodity code'),
+        }
+
+    def get_context_data(self, **kwargs):
+        rules_regs = None
```
richtier: rules_regulation` would be more apt

core/views.py
```diff
@@ -1,4 +1,41 @@
...
+        if self.request.GET.get('country'):
+            rules_regs = helpers.get_rules_and_regulations(self.request.GET.get('country'))
+        return super().get_context_data(rules_regs=rules_regs, **kwargs)
+
+    def form_valid(self, form):
+        rules_regs = helpers.get_rules_and_regulations(self.request.GET.get('country'))
+        helpers.create_export_plan(sso_session_id=self.request.user.session_id, exportplan_data=rules_regs)
+        self.success_url = reverse_lazy('core:exportplan-view')
```
richtier: instead of overwriting `success_url` `get_success_url` is preferable

core/views.py
```diff
@@ -1,4 +1,41 @@
...
+        return super().form_valid(form)
+
+
+class ExportPlanView(TemplateView):
+    template_name = 'ExportPlanView.html'
```
richtier: lower case template names please

core/templates/ExportPlanView.html
```diff
@@ -0,0 +1,35 @@
+{% load static from staticfiles %}
```
richtier: lower case template names please

also please move to `core/templates/core/` so in views we can do `template_name = 'core/foo``

PR #31, opened 2020-02-18
Company questions modal

richtier:

 - [x] Changelog entry added.
 
progress


core/sass/_dashboard.scss
```diff
@@ -0,0 +1,79 @@
+@import 'helpers/_colours';
```
eadpearce: don't need to put the `_` in import paths

PR #36, opened 2020-02-24
Mvp 228 duties url

depsiatwal:
 - [x] Changelog entry added.

### DIFF
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -150,3 +150,7 @@ def get_dashboard_export_opportunities():
             'closing_data': '11 March 2020',
         }
     ]
+
+
+def get_custom_duties_url(product_code, country):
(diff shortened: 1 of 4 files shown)


core/helpers.py
```diff
@@ -150,3 +150,7 @@ def get_dashboard_export_opportunities():
             'closing_data': '11 March 2020',
         }
     ]
+
+
+def get_custom_duties_url(product_code, country):
+    return f'https://www.check-duties-customs-exporting-goods.service.gov.uk/summary?d={country}&pc={product_code}'
```
kowalcj0: It'd be useful to get the domain name from settings. 
As it looks like that soon, there should be a working Staging & UAT env.

PR #39, opened 2020-03-02
Add Lesson model

WillTaylor22:
 - [x] Changelog entry added.

lesson


core/urls.py
```diff
@@ -47,3 +48,9 @@ def anonymous_user_required(function):
         name='api-create-company'
     ),
 ]
+
+urlpatterns += [
+    path('lessons/', learn_views.lesson_list, name='lesson_list'),
+    path('lesson/<int:id>/', learn_views.lesson_details, name='lesson_details'),
+    path('lesson/<int:id>/delete/', learn_views.lesson_delete, name='lesson_delete'),
+]
```
sdonk: Should these lines be wagtail url entries?

learn/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.contrib import admin
+
+# Register your models here.
+
+from learn.models import Lesson
+admin.site.register(Lesson)
```
sdonk: Wagtail doesn't need explicit admin model registering

learn/models.py
```diff
@@ -0,0 +1,29 @@
...
+        Topic,
+        models.SET_NULL,
+        blank=True,
+        null=True,)
+    position = PositionField(collection='topic')
+    update_date = models.DateTimeField(auto_now=True)
+    create_date = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        ordering = ['-id']
```
sdonk: Should these be wagtail models?
Maybe orderable https://docs.wagtail.io/en/v2.8/reference/pages/model_reference.html#wagtail.core.models.Orderable

learn/views.py
```diff
@@ -0,0 +1,52 @@
+from django.shortcuts import render, get_object_or_404
+from django.middleware.csrf import get_token
+from django.http import JsonResponse
+from learn.models import Lesson
+from learn.forms import LessonForm
+
+
+def lesson_list(request, template='lesson/list.html'):
```
sdonk: Should these be wagtail views? In that case you don't need the form class either

core/urls.py
WillTaylor22: Wagtail sounds like a good way to proceed. This was initially meant to be something quick and easy to test the backend with, but it makes sense that we'd want to use it in Wagtail.

WillTaylor22: And additionally, pre-designs, to have a simple user-facing page to work with.

PR #41, opened 2020-03-02
Add export plan skeleton

eadpearce:

 - [x] Changelog entry added.

FireShot Capture 029 - Welcome to great gov uk - greatcms trade great


config/urls.py
```diff
@@ -18,7 +19,8 @@
     path('admin/', include(wagtailadmin_urls)),
     path('documents/', include(wagtaildocs_urls)),
     path('sso/', include(sso.urls)),
-    path('', include(core.urls)),
+    path('', include(core.urls, namespace="core")),
+    path('export-plan/', include(exportplan.urls, namespace="exportplan")),
```
sdonk: Please don't mix single and double quotes :)

PR #42, opened 2020-03-03
MVP-225 add local time

depsiatwal:
 - [x] Changelog entry added.


core/helpers.py
```diff
@@ -155,3 +158,21 @@ def get_dashboard_export_opportunities():
...
 def get_custom_duties_url(product_code, country):
     return f'{settings.MADB_URL}/summary?d={country}&pc={product_code}'
+
+
+def country_code_iso3_to_iso2(iso3_country_code):
+    if countries_by_alpha3.get(iso3_country_code):
+        return countries_by_alpha3['CHN'].alpha2
+
+
+def get_local_time(country_code):
```
sdonk: Can we use any of the Django boilerplate? https://docs.djangoproject.com/en/3.0/topics/i18n/timezones/#localtime

core/helpers.py
```diff
@@ -155,3 +158,21 @@ def get_dashboard_export_opportunities():
...
+    local_time = {}
+    format = '%Y-%m-%d %H:%M %Z%z'
+    iso3_country_code = country_code_iso3_to_iso2(country_code)
+    if iso3_country_code and pytz.country_timezones('CN'):
+        local_time["timezone"] = pytz.country_timezones('CN')[0]
```
sdonk: Please don't mix single and double quotes :)

tests/unit/core/test_helpers.py
```diff
@@ -200,3 +201,20 @@ def test_create_user_profile_failure(mock_create_user_profile, user, rf):
...
+
+def test_country_code_iso3_to_iso2():
+    assert helpers.country_code_iso3_to_iso2('CHN') == 'CN'
+
+
+def test_country_code_iso3_to_iso2_not_found():
+    assert helpers.country_code_iso3_to_iso2('XNY') is None
+
+
+@freeze_time("2012-01-14 03:21:34")
```
sdonk: Please don't mix single and double quotes :)

kowalcj0: https://pre-commit.com/ with flake8 & black FTW
and possibly switch to double quotes -> https://github.com/psf/black#strings :)

sdonk: https://www.python.org/dev/peps/pep-0008/#string-quotes

PEP8 doesn't express any preference. The codebase has mostly single quotes, so we should keep using those for consistency :) 
Also, for double you have to press shift as well as the other key which is the reason why flake8 package has got single quotes as preference

sdonk: The flake8 job will not fail if the incorrect quotes are used

PR #43, opened 2020-03-03
implement flake8 rules

sdonk:


sdonk:
Not sure why codecov is complaining...

PR #44, opened 2020-03-03
Implement design for login page

richtier:
 - [x] Changelog entry added.
 
login


config/settings.py
```diff
@@ -418,3 +418,6 @@
 MADB_URL = env.str(
     'MADB_URL', 'https://www.check-duties-customs-exporting-goods.service.gov.uk'
 )
+
+# directory constants
+DIRECTORY_CONSTANTS_URL_SINGLE_SIGN_ON = env.str('DIRECTORY_CONSTANTS_URL_SINGLE_SIGN_ON', '')
```
sdonk: Is this longer than 119? :)

core/context_processors.py
```diff
@@ -10,5 +10,6 @@ def signup_modal(request):
         'linkedin_url': add_next(settings.SSO_OAUTH2_LINKEDIN_URL, url),
         'google_url':  add_next(settings.SSO_OAUTH2_GOOGLE_URL, url),
         'terms_url': urls.domestic.TERMS_AND_CONDITIONS,
+        'password_reset_url': urls.domestic.SINGLE_SIGN_ON / 'accounts/password/reset/',
```
sdonk: Is this longer than 119? :)

config/settings.py
richtier: nope :) it's around 90

core/context_processors.py
richtier: nope :)

PR #46, opened 2020-03-05
MVP256 add dataservices

depsiatwal:
 - [x] Changelog entry added.


core/helpers.py
```diff
@@ -168,3 +168,19 @@ def get_timezone(country_code):
...
+
+def get_exportplan_marketdata(country_code):
+    # This is a temp wrapper for MVP as we finalise the source(s) this should move to backend
+    country_code = 'CHN'
+    exportplan_marketdata = {}
+    exportplan_marketdata['timezone'] = get_timezone(country_code)
+    response = api_client.dataservices.get_corruption_perceptions_index(country_code)
+    response.raise_for_status()
+    if response.json():
+        exportplan_marketdata['corruption_perceptions_index'] = response.json()
```
richtier: normally there is no need to do `if response.json()`, we can just do `        exportplan_marketdata['corruption_perceptions_index'] = response.json()
`

if we do need to check if the json is not empty then please do something like 

```
parsed = response.json()
if parsed:
    ...
```

core/helpers.py
```diff
@@ -168,3 +168,19 @@ def get_timezone(country_code):
...
+    response = api_client.dataservices.get_easeofdoingbusiness(country_code)
```
richtier: please avoid overwriting the previous `response` - it makes debugging in pdb and sentry harder.

something like:
```
response_corruption_perceptions_index = 
response_ease_of_doing_business = 
```

PR #57, opened 2020-03-12
Mvp-310 add comtrade data

depsiatwal:

 - [x] Changelog entry added.


config/settings.py
```diff
@@ -412,3 +412,4 @@
 
 # directory constants
 DIRECTORY_CONSTANTS_URL_SINGLE_SIGN_ON = env.str('DIRECTORY_CONSTANTS_URL_SINGLE_SIGN_ON', '')
+DIRECTORY_CLIENT_CORE_CACHE_EXPIRE_SECONDS = 3600
```
richtier: please avoid magic numbers. `60 * 60 * 30  # 30 days` is easier to understand

PR #60, opened 2020-03-16
Add sector chooser component

eadpearce:

 - [x] Changelog entry added.

Screenshot 2020-03-16 at 14 14 41
Screenshot 2020-03-16 at 14 14 47
Screenshot 2020-03-16 at 14 15 00
Screenshot 2020-03-16 at 14 15 13


exportplan/urls.py
```diff
@@ -5,8 +5,9 @@
 app_name = 'exportplan'
 
 urlpatterns = [
-    path('about-your-business/', views.ExportPlanBuilderSectionView.as_view(), name='about-your-business'),
     path('start/', views.ExportPlanStartView.as_view(), name='start'),
     path('create/', views.ExportPlanCreateView.as_view(), name='create'),
-
+    path('', views.ExportPlanLandingPageView.as_view(), name='landing-page'),
+    path('dashboard/', views.ExportPlanDashboardView.as_view(), name='dashboard'),
```
richtier: I guess this was a merge conflict

previous PR moved the dashboard page into the CMS, this moves it back out again :)

react-components/src/components/Sector.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { slugify } from '../Helpers'
+
+
+export default class Sector extends React.Component {
```
richtier: looks like this could be a simple function

if using `React.Component` so you can use state then consider react hooks `useState`

react-components/src/components/SectorChooser.jsx
```diff
@@ -0,0 +1,163 @@
...
+    this.handleMouseOut = this.handleMouseOut.bind(this)
+  }
+
+  addRemoveSector = (sector) => {
+    if (this.state.selectedSectors.indexOf(sector) == -1) {
+      this.setState({selectedSectors: this.state.selectedSectors.concat([sector])})
+      return true
+
+    } else {
+      const updatedSelectedSectors = this.state.selectedSectors.filter(
```
richtier: could use terser syntax
`const updatedSelectedSectors = this.state.selectedSectors.filter(item => item != sector)`

PR #64, opened 2020-03-17
bugfix/remove-reference-to-sectors

WillTaylor22:
Same as previous changelog entry really

### DIFF
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -75,7 +75,7 @@ def get_dashboard_events(sso_session_id):
 
 def get_dashboard_export_opportunities(company, sso_session_id):
     sectors = company and company.expertise_industries_labels
-    search_term = ' '.join(sectors)
+    search_term = ' '.join(sectors) if sectors else ''
     results = api_client.personalisation.export_opportunities_by_relevance_list(sso_session_id, search_term)
     if (results.status_code == 200):
         return parse_opportunities(results.json()['results'])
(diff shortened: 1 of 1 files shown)


core/helpers.py
```diff
@@ -74,8 +74,8 @@ def get_dashboard_events(sso_session_id):
 
 
 def get_dashboard_export_opportunities(company, sso_session_id):
-    sectors = (company and company.expertise_industries_labels) or list(CompanyParser.SECTORS.values())
-    search_term = ' '.join(sectors)
+    sectors = company and company.expertise_industries_labels
+    search_term = ' '.join(sectors) if sectors else ''
```
sdonk: You don't need the if statement

`In [6]: ' '.join([])
Out[6]: ''`

core/helpers.py
```diff
@@ -74,8 +74,8 @@ def get_dashboard_events(sso_session_id):
 
 
 def get_dashboard_export_opportunities(company, sso_session_id):
-    sectors = (company and company.expertise_industries_labels) or list(CompanyParser.SECTORS.values())
```
richtier: the goal of this line was _use the user's industries of interest if provided otherwise fallback to the list of all industries_

if we remove the _otherwise use the list of all industries_ then we're not handling the case when users have not yet specified the industries of interest.

sdonk:
Is this the same of https://github.com/uktrade/great-cms/pull/73 @WillTaylor22

core/helpers.py
WillTaylor22: That's already achieved. if blank the endpoint returns everything.

PR #65, opened 2020-03-17
Add create_pages command

richtier:


Procfile
```diff
@@ -1 +1 @@
-web: python manage.py collectstatic --noinput && python manage.py distributed_migrate --noinput && gunicorn config.wsgi --bind 0.0.0.0:$PORT
+web: python manage.py collectstatic --noinput && python manage.py distributed_migrate --noinput && python manage.py create_pages && gunicorn config.wsgi --bind 0.0.0.0:$PORT
```
sdonk: Uhm, why running this command everytime we run the webserver?

richtier: so "required" pages are guaranteed to exist. Note that the factory will perform `get_or_create` because of the change I made to factory Meta

sdonk: We use the Procfile on PaaS only and once the pages are created they'll be there. 
It's a one off task that doesn't need to be run every time the app starts on PaaS.

I think this needs to be removed from here :)

PR #71, opened 2020-03-20
Add export plan dummy pages

eadpearce:

 - [ ] Changelog entry added.

FireShot Capture 031 - Welcome to great gov uk - greatcms trade great


exportplan/data.py
```diff
@@ -12,3 +15,12 @@
...
     'Action plan',
 ]
+
+SECTION_SLUGS = [
+    slugify(section) for section in SECTION_TITLES
+]
+
+SECTION_URLS = [
+    reverse_lazy('exportplan:section', kwargs={'slug': slug})
+    for slug in SECTION_SLUGS
```
sdonk: can this go on one line?

eadpearce: oh sure

exportplan/views.py
```diff
@@ -20,8 +20,32 @@ def get_context_data(self, *args, **kwargs):
...
-class ExportPlanBuilderSectionView(BaseExportPlanView):
-    template_name = 'exportplan/builder_section.html'
+class ExportPlanSectionView(BaseExportPlanView):
+
+    @property
+    def slug(self, **kwargs):
+        return self.kwargs['slug']
+
+    @property
+    def template_name(self, **kwargs):
```
richtier: can you use the built-in django view `get_template_name` pls 

http://ccbv.co.uk/projects/Django/3.0/django.views.generic.base/TemplateView/#get_template_names

exportplan/views.py
```diff
@@ -20,8 +20,32 @@ def get_context_data(self, *args, **kwargs):
...
+        return f'exportplan/sections/{self.slug}.html'
+
+    @property
+    def next_section(self):
+        index = data.SECTION_SLUGS.index(self.slug)
+
+        if index == len(data.SECTION_SLUGS) - 1:
```
richtier: could instead do `if self.slug == data.SECTION_SLUGS[-1]`

I think that better communicates to the reader the purpose: _if the slug is the last in the list..._

makefile
```diff
@@ -7,7 +7,7 @@ clean:
 ENV_FILES?='test,dev'
 pytest:
 	ENV_FILES=$(ENV_FILES) \
-	pytest tests/unit $(ARGUMENTS)
+	pytest tests/unit --cov=. --cov-config=.coveragerc --cov-report=html $(ARGUMENTS)
```
richtier: this will run codecov locally. is that desired?

it's currently done only in circle https://github.com/uktrade/great-cms/blob/develop/.circleci/config.yml#L29

tests/unit/exportplan/test_views.py
```diff
@@ -102,3 +96,9 @@ def test_exportplan_view(
     assert response.context['utz_offset'] == '+0800'
     assert response.context['lastyear_import_data'] == {'last_year_data_partner': {'Year': 2019, 'value': 10000}}
     assert response.context['historical_import_data'] == {'historical_data_all': {'Year': 2019, 'value': 1234}}
+
+
+@pytest.mark.parametrize('url', data.SECTION_URLS)
```
richtier: :heart:

exportplan/views.py
eadpearce: oh i see. i forgot you can use negative indexing in python

makefile
eadpearce: yeah i wanted to run it so i can check locally what is covered before pushing. the htmlcov directory is already in the gitignore so it won't get committed

exportplan/views.py
eadpearce: forgot about this. ty

PR #73, opened 2020-03-20
Chore/change format of exops api helper 20 mar

WillTaylor22:
No changelog item as v. minor bugfix


core/helpers.py
```diff
@@ -73,9 +73,9 @@ def get_dashboard_events(sso_session_id):
     return []
 
 
-def get_dashboard_export_opportunities(company, sso_session_id):
-    sectors = (company and company.expertise_industries_labels) or list(CompanyParser.SECTORS.values())
-    search_term = ' '.join(sectors)
+def get_dashboard_export_opportunities(sso_session_id, company):
+    sectors = (company and company.expertise_industries_labels)
+    search_term = ' '.join(sectors) if sectors else ''
```
sdonk: you don't need the `if sectors else ''` part, `join` returns an empty string if sectors is None

WillTaylor22: Are you 100%?

```
>>>  ' '.join(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can only join an iterable

```

sdonk: Sorry, I meant empty iterable

`
In [1]: ' '.join([])
Out[1]: ''
`

PR #80, opened 2020-03-24
MVP-356 - show target-market data

depsiatwal:
 - [x] Changelog entry added.


exportplan/templates/exportplan/sections/target-markets.html
```diff
@@ -31,7 +31,58 @@ <h1 class="h-xl">Target markets</h1>
 </section>
 
 <hr>
+{% if export_marketdata %}
+  {% load humanize %}
```
richtier: treat these like python imports - so they should live at the top of file :)

also please be specific in what's being imported. right now it's equivelant to `from foo import *`, which is hard to understand whats coming from where. please instead do `{% load intcomma from humanize %}`

exportplan/views.py
```diff
@@ -3,18 +3,44 @@
...
 
 from directory_constants.choices import INDUSTRIES
 
 from exportplan import data, forms, helpers
 
 
-class BaseExportPlanView(TemplateView):
+class ExportPlanDataMixin:
+    def get_context_data(self, *args, **kwargs):
+        if self.request.path == reverse('exportplan:section', kwargs={'slug': 'target-markets'}):
```
richtier: it looks like this should not be in a mixin used in a base view because this is checking if a specific view is being accessed (`reverse('exportplan:section', kwargs={'slug': 'target-markets'})`). 

should this be moved to the `get_context_data` of that specific view rather than added to the base class?

exportplan/views.py
```diff
@@ -3,18 +3,44 @@
...
+            rules_regulation = helpers.get_exportplan_rules_regulations(sso_session_id=self.request.user.session_id)
+            if rules_regulation:
+                export_marketdata = helpers.get_exportplan_marketdata(rules_regulation.get('country_code'))
+                utz_offset = datetime.now(
+                    pytz.timezone(export_marketdata['timezone'])).strftime('%z')
```
richtier: could be one line

PR #82, opened 2020-03-25
MVP-344 Add wagtail cache

sdonk:
 - [x] Changelog entry added.


config/settings.py
```diff
@@ -116,7 +119,7 @@
     REDIS_URL = env.str('REDIS_URL')
 
 
-if env.bool('API_CACHE_DISABLED', False):
+if env.bool('CACHE_DISABLED', False):
```
sdonk: There's no api anymore, the name was confusing

PR #87, opened 2020-03-27
MVP-362 - improve UI on country statistics view

Trunk89:
 - [x] Changelog entry added.

- Updated the HTML to be semantic
- Moved HTML into a separate partial
- Moved buttons stylings into a common space

Updated view


core/sass/typography/_text.scss
```diff
@@ -0,0 +1,3 @@
+.text--visually-red {
```
eadpearce: we already have a class for this! it's `text-flag-red`

Trunk89: Nice! I will make sure to update it when fixing the merge conflicts

PR #90, opened 2020-03-31
Implement new signup modal design

richtier:

 - [x] Changelog entry added.
 
signup


react-components/src/views/SignupModal/StepCredentials.jsx
```diff
@@ -5,17 +5,35 @@ import Services from '@src/Services'
...
 import SocialLoginButtons from '@src/components/SocialLoginButtons'
 
-import './stylesheets/Step1.scss'
+import './stylesheets/StepCredentials.scss'
 
 
-export default function Step1(props){
+export default function StepCredentials(props){
+
+  function getLede() {
```
depsiatwal: do we need more meaningful name here ?

PR #91, opened 2020-04-01
MVP-365 - Tooltip functionality

Trunk89:

 - [ x ] Change has a jira ticket that has the correct status.
 - [ x ] Changelog entry added.
 - added educational moment tooltip functionality 
 - added a dynamic way of instantiating common components from HTML selectors
 - corrected the webpack img loader link for jest tests
 - added asset copy task for assets needed by CSS as they are not automatically moved to dist folder by React
 - added `npm run test:dev` task to run tests with watch option on local env

Tooltip


react-components/src/bundle.js
```diff
@@ -6,10 +6,14 @@ import CountriesModal from '@src/views/CountriesModal/Base'
 import LoginModal from '@src/views/LoginModal/Modal'
 import Tour from '@src/views/Tour/Base'
 import { createSectorChooser } from '@src/components/SectorChooser'
-import {STEP_CREDENTIALS, STEP_VERIFICATION_CODE} from '@src/views/SignupModal/Wizard/'
+import { STEP_CREDENTIALS, STEP_VERIFICATION_CODE } from '@src/views/SignupModal/Wizard/'
```
jestanoff, line 9: We need Eslint/Prettier :)

Trunk89: I was thinking the same :D Will add it as a separate story

react-components/src/bundle.js
```diff
@@ -13,6 +13,10 @@ import { STEP_CREDENTIALS, STEP_VERIFICATION_CODE } from '@src/views/SignupModal
 import Services from '@src/Services'
 import '@babel/polyfill'
 
+//initialise common components automatically based on selectors
+import ComponentInitialiser from './initialisers/common.initialiser'
+new ComponentInitialiser();
```
richtier, line 18: This causes including the js bundle on a page to have an  implicit side-effects: activating the tooltips. I think this is an anti-pattern because
- implicit execution can be hard to debug. Especially if we treat implicit execution as the new "way to do it" and implicitly execute more components as we add new ones.
- monolithic: cannot include MVPdit.js without executing that `ComponentInitialiser`. That can bite us.
- if executing `ComponentInitialiser` fails then all the js that would execute after it will breaks.

Can solve by moving to here: https://github.com/uktrade/great-cms/blob/develop/core/templates/core/base.html#L36

e.g,

```
{% block body_js %}
    {{ block.super }}
    new ComponentInitialiser();
    ...
{% endblock %}
```

as that would be called in `<body>` that will also remove the need for `DOMContentLoaded` listener so reduces complexity.

react-components/src/components/EducationalMomentIcon/EducationalMomentIcon.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import './EducationalMomentIcon.scss';
+import OpenBookIcon from '../../../assets/open-book-icon.png';
```
richtier, line 3: import OpenBookIcon from '@assets/open-book-icon.png' (alias defined in webpack.config)

react-components/src/components/EducationalMomentIcon/EducationalMomentIcon.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import './EducationalMomentIcon.scss';
```
richtier, line 2: Note there is need to end the line with semi colon as Babel adds them

http://www.ecma-international.org/ecma-262/7.0/index.html#sec-rules-of-automatic-semicolon-insertion

> Most ECMAScript statements and declarations must be terminated with a semicolon. Such semicolons may always appear explicitly in the source text. For convenience, however, such semicolons may be omitted from the source text in certain situations. **These situations are described by saying that semicolons are automatically inserted into the source code token stream in those situations.**

ending in semi colon does no harm of course, just less faff :)

react-components/src/services/components.service.js
```diff
@@ -0,0 +1,26 @@
+let initialisedComponents = {};
```
richtier, line 1: could you also move `Services.js` to `services` too please as it may be confusing having `Services.js` as a sibling of `services` folder. Also some OSs can be configured to be case insensitive file system :scream: so don't want to cause unexpected outcomes :)

react-components/src/initialisers/common.initialiser.js
```diff
@@ -0,0 +1,38 @@
...
+
+        document.addEventListener('dom-change', (event) => {
+            this.initialiseComponents(event.target, COMPONENTS_CONFIG, COMPONENTS);
+        });
+    }
+
+    initialiseComponents(doc, config, components) {
+        config.forEach((component) => {
+            if (!ComponentService.getInitialisedComponent(component.name)) {
+                ComponentService.removeInitialisedComponents(component.name)
```
richtier, line 22: hmm is that correct? 
```
if component has _not_ been initialized 
then remove it
```

I would expect it to be 

 
```
if component _has_ been initialized 
then remove it
```

react-components/src/services/components.service.js
```diff
@@ -0,0 +1,26 @@
...
+    static getInitialisedComponents() {
+        return initialisedComponents;
+    }
+
+    static getInitialisedComponent(component) {
+        return initialisedComponents[component];
+    }
+
+    static removeInitialisedComponents(component) {
+        initialisedComponents[component] = [];
```
richtier, line 20: this does not destroy the initialised component - just removes reference to it. Should it?

PR #94, opened 2020-04-03
Hook up country data to FE

eadpearce:

 - [x] Changelog entry added.
Screenshot 2020-04-03 at 17 20 36


kowalcj0:
In order to fix failing tests and fix merge conflicts I had to modify few things, like:
* strip white spaces from MADB country list `get_madb_country_list()`
* response handling on 200 OK in `Services.js`
* fix position of customisation bar on export-plan/dashboard/ as it was causing trouble with page tour

PR #111, opened 2020-04-14
Mvp 306 countries

depsiatwal:
 - [ ] Changelog entry added.


exportplan/ajax.py
```diff
@@ -35,3 +33,31 @@ def get(self, request):
...
         return JsonResponse(data)
+
+
+class ExportPlanRecommendedCountriesDataView(views.APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+
+        if not self.request.GET.get('sectors'):
+            return HttpResponse(status=400)
```
richtier, line 44: we can use a serializer class to reduce codez:

```
# serializers.py
class ExportPlanRecommendedCountriesSerializer(serializers.Serializer):
    sectors = serializers.ListField(child=serializers.CharField())

class ExportPlanRecommendedCountriesDataView(views.APIView):
    permission_classes = [IsAuthenticated]
    serializer_class = serializers.ExportPlanRecommendedCountriesSerializer
 
    def get(self, request):
        serializer = self.serializer_class(data=request.data)
        serializer.is_valid(raise_exception=True)  # this causes the view to automatically return a 400 response with the serializer's errors in the body of the response.
       ...
       sectors = serializer.validated_data['sectors']   # this will be a list of strings already - no need to call `sectors.split(',')`

File: exportplan/ajax.py [hunk 2]
```diff
@@ -35,3 +33,31 @@ def get(self, request):
...
+
+        sectors = self.request.GET.get('sectors')
+
+        try:
+            # To make more efficient by removing get export plan
+            export_plan = helpers.get_exportplan(sso_session_id=self.request.user.session_id)
+            helpers.update_exportplan(
+                sso_session_id=self.request.user.session_id,
+                id=export_plan['pk'],
+                data={'sectors': sectors.split(',')}
```
richtier, line 54: if did the serializer thing above then this could be `data=serializer.validated_data` :sunglasses:

exportplan/ajax.py
```diff
@@ -35,3 +33,31 @@ def get(self, request):
...
+            )
+            recommended_countries = helpers.get_recommended_countries(
+                sso_session_id=self.request.user.session_id,
+                sectors=sectors
+            )
+        except ReadTimeout:
+            return HttpResponse(status=504)
```
richtier, line 61: I think it's better to not catch ReadTimeout and instead just let the orig exception just happen - as there will be more info logged to sentry (right now we're returning 50x response and _not_ logging anything in sentry so devs will get no sentry alert that there are problems afoot happened.

exportplan/ajax.py
```diff
@@ -35,3 +33,31 @@ def get(self, request):
...
+        data = {'countries': recommended_countries, }
+        return JsonResponse(data)
```
richtier, line 63: as this is a DRF view, we can return a good old simple `rest_framework.response.Response`. DRF defaults to it being JSON content type

exportplan/helpers.py
```diff
@@ -94,3 +94,15 @@ def get_comtrade_historicalimportdata(commodity_code, country):
...
+
+
+def get_recommended_countries(sso_session_id, sectors):
+    response = api_client.personalisation.recommended_countries_by_sector(sso_session_id=sso_session_id, sector=sectors)
+    response.raise_for_status()
+    parsed = response.json()
+    if parsed:
+        for recommended_country in parsed:
+            country = recommended_country['country'].title()
+            recommended_country['country'] = country
```
richtier, line 106: if we had a shorter variable name then we could make it a bit easier to read:

        for item in parsed:
            item['country'] = item['country'].title()

```

File: exportplan/urls.py [hunk 6]
```diff
@@ -9,5 +9,10 @@
     path('create/', views.ExportPlanCreateView.as_view(), name='create'),
     path('section/target-markets/', views.ExportPlanTargetMarketsView.as_view(), name='target-markets'),
     path('section/<slug:slug>/', views.ExportPlanSectionView.as_view(), name='section'),
+    path(
+        'ajax/recommended-countries/',
```
richtier, line 13: to be consistent with other API urls can this be something like`api/recommended-countries/` pls

react-components/src/components/RecommendedCountries.jsx
```diff
@@ -0,0 +1,51 @@
...
+export default class RecommendedCountries extends React.Component {
+  constructor(props) {
+    super(props)
+    this.countryRef = React.createRef()
+  }
+
+  isSelectedAlready(countryName) {
+    const { selectedCountries } = this.props
+
+    return selectedCountries.filter((country) => country.country === countryName).length > 0
```
richtier, line 15: `return selectedCountries.some(country => country.country === countryName)`

pretty cool - it fails out of the check as soon as at least one return true - so a little quicker. slightly easier to read too I think

exportplan/ajax.py
eadpearce: This was mainly added as timeouts were happening a LOT and it was nice to have it passed to the react component. If there's a way to still do that and send it to sentry that would be good. @depsiatwal probably knows more about this though

PR #113, opened 2020-04-16
Use business sso user on non-admin pages

richtier:


config/settings.py
```diff
@@ -344,15 +344,16 @@
 
 ENFORCE_STAFF_SSO_ENABLED = env.bool('ENFORCE_STAFF_SSO_ENABLED', False)
 
+# authbroker config
+AUTHBROKER_URL = env.str('STAFF_SSO_AUTHBROKER_URL')
+AUTHBROKER_CLIENT_ID = env.str('AUTHBROKER_CLIENT_ID')
+AUTHBROKER_CLIENT_SECRET = env.str('AUTHBROKER_CLIENT_SECRET')
```
richtier, line 350: we import the backend to subclass it. these settings are used at module of the imported file, so these need to be present in settings

PR #122, opened 2020-04-20
MVP-405 - create export plan using personalisation data

depsiatwal:
 - [x] Changelog entry added.


exportplan/helpers.py
```diff
@@ -106,3 +106,28 @@ def get_recommended_countries(sso_session_id, sectors):
...
+        target_markets = target_markets + [{'country': c} for c in user.company.expertise_countries_labels]
+    return {
+        'export_countries': [rules_regulations['country']],
+        'export_commodity_codes': [rules_regulations['commodity_code']],
+        'rules_regulations': rules_regulations,
+        'target_markets': target_markets,
+    }
+
+
+def get_export_plan_or_create(user):
```
richtier, line 123: naming is hard. let's be inspired by django: `get_or_create_export_plan`  - similar to Model.objects.get_or_create` :)

exportplan/helpers.py
```diff
@@ -106,3 +106,28 @@ def get_recommended_countries(sso_session_id, sectors):
...
+    # This is a temp hook to create initial export plan. Once we have a full journey this can be removed
+    export_plan = get_exportplan(user.session_id)
+
+    if not export_plan:
+        rules = get_rules_and_regulations('Australia')
```
richtier, line 128: hard coded to Australia on purpose?

exportplan/views.py
```diff
@@ -22,6 +22,10 @@ def get_context_data(self, *args, **kwargs):
             country_choices=json.dumps(country_choices),
             *args, **kwargs)
 
+    def dispatch(self, *args, **kwargs):
+        self.export_plan = helpers.get_export_plan_or_create(self.request.user)
```
richtier, line 26: this adds complexity for the reader but does not seem to add value compared with just calling `helpers.get_export_plan_or_create(self.request.user)` in `get_context_data` (where we're currently using `self.export_plan`

exportplan/helpers.py
depsiatwal: Yes this is hardcoded since this is the core company that appears on top. Export plan we were told doesn't need to allow for multiple countries and commodity codes for MVP.

PR #125, opened 2020-04-21
MVP-406 - update allow remove export plan

depsiatwal:
 - [x] Changelog entry added.


exportplan/ajax.py
```diff
@@ -36,6 +36,34 @@ def get(self, request):
         return JsonResponse(data)
 
 
+class ExportPlanRemoveCountryDataView(views.APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        if not self.request.GET.get('country'):
+            return HttpResponse(status=400)
```
richtier, line 44: please use a serializer for this, less custom code

exportplan/ajax.py
```diff
@@ -36,6 +36,34 @@ def get(self, request):
...
+            export_plan = helpers.get_exportplan(sso_session_id=self.request.user.session_id)
+            target_markets_filtered = filter(lambda i: i['country'] != country, export_plan['target_markets'])
+            data = list(target_markets_filtered)
+            export_plan = helpers.update_exportplan(
+                sso_session_id=self.request.user.session_id,
+                id=export_plan['pk'],
+                data={'target_markets': data}
+            )
+        except ReadTimeout:
+            return HttpResponse(status=504)
```
richtier, line 57: this hides the error from devs. if we did not catch the error then it would also be logged in sentry. currently it returns 5xx to user but we're blind to it happening

exportplan/ajax.py
```diff
@@ -36,6 +36,34 @@ def get(self, request):
...
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        if not self.request.GET.get('country'):
+            return HttpResponse(status=400)
+        country = self.request.GET.get('country')
+        try:
+            # To make more efficient by removing get
+            export_plan = helpers.get_exportplan(sso_session_id=self.request.user.session_id)
+            target_markets_filtered = filter(lambda i: i['country'] != country, export_plan['target_markets'])
```
richtier, line 49: list comprehension is more readable:

```
data = [item for item  in export_plan['target_markets'] if item['country'] != country]

```

that also removes the needs to coerce to list in the line after - as it's already a list

react-components/src/Services.js
```diff
@@ -43,6 +43,11 @@ const getCountryData = function (country) {
   return get(config.countryDataUrl, { country: country }).then((response) => responseHandler(response).json())
 }
 
+const removeCountryData = function (country) {
+  console.log('config.removeCountryDataUrl')
```
richtier, line 47: please remove debugging lines

PR #126, opened 2020-04-21
MVP-407 refactor browser tests

kowalcj0:
What happened here:
* all Browser related fixtures moved to tests/browser/conftest.py
* all common steps moved to tests/browser/steps.py
* got rid of duplicated fixtures
* Browser test code reformatted with isort & black
* added configs for [isort](https://github.com/timothycrosley/isort) & [pre-commit](https://pre-commit.com/)
* removed redundant DB setup step for both unit & browser tests [as these two don't need to run migrations and create dedicated test_greatcms DB (as pytest is configured with `--nomigrations --reuse-db`)]

 - [x] Changelog entry added.


WillTaylor22:
LGTM

PR #140, opened 2020-04-28
Fix country data percentages

eadpearce:

 - [x] Changelog entry added.


jestanoff:
Looks good to go

PR #154, opened 2020-05-01
Prevent learn caurosel twice

richtier:
 - [x] Changelog entry added.
 -
learn-redirect

### DIFF
--- a/makefile
+++ b/makefile
@@ -40,7 +40,7 @@ check_migrations:
 	yes n | ENV_FILES=$(ENV_FILES) ./manage.py migrate --plan
 
 webserver:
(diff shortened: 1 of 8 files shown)


makefile
```diff
@@ -65,8 +65,8 @@ test_load:
 	-$(kill_webserver)
 
 requirements:
-	pip-compile -r --annotate requirements.in
-	pip-compile -r --annotate requirements_test.in
+	pip-compile --upgrade -r --annotate requirements.in
```
richtier, line 68: turns out this flag is the secret source to make the abstract version numbers do their magic

PR #155, opened 2020-05-01
MVP-420 tests for tool-tips on target markets data page

kowalcj0:
 - [x] Changelog entry added.


README.md
```diff
@@ -60,6 +60,7 @@ Secrets such as API keys and environment specific configurations are placed in `
 | make boostrap_great           | Create required database records so the CMS works |
 | make create_pages             | Create required Pages |
 | make create_tours             | Create required Page tours |
+| make recreate                 | Runs: database, boostrap_great, create_pages, create_tours in one go |
```
richtier, line 63: good idea. naming is hard tho, to the point I think `recreate` is not  great name, but I cant think of a btter one :laughing:

tests/browser/common_selectors.py
```diff
@@ -26,6 +26,16 @@ def is_visible(self):
     def is_authenticated(self):
         return self.value.is_authenticated
 
+    @property
+    def selector_template(self):
+        return self.value.selector_template
+
+    def __str__(self) -> str:
+        return self.value.name if (hasattr(self.value, 'name') and self.value.name) else self.__class__.__name__
```
richtier, line 34: could this be `getattr(self.value, 'name', None) or self.__class.__.__name__`?

tests/browser/common_selectors.py
```diff
@@ -26,6 +26,16 @@ def is_visible(self):
...
+
+    def __repr__(self) -> str:
+        return self.value.name if (hasattr(self.value, 'name') and self.value.name) else self.__class__.__name__
+
```
richtier, line 38: I think we can do `__repr__= __str__`

tests/browser/common_selectors.py
```diff
@@ -43,37 +53,66 @@ class ElementType(Enum):
...
+        return super(Selector, cls).__new__(
+            cls,
+            by,
+            selector=selector,
+            type=type,
+            is_visible=is_visible,
+            is_authenticated=is_authenticated,
+            name=name,
+            selector_template=selector_template,
+        )
```
richtier, line 85: this is very verbose and does some things many devs will get at least a little puzzled by: using `__new__` instead of __init__`, executing a named tuple "inline" as the class is being defined.

in terms of readability I think this is improved:

```
from typing import NamedTuple

class Selector(NamedTuple):
    by: By
    selector: str
    type: ElementType = None
    is_visible: bool = True
    is_authenticated: bool = False
    name: str = None
    selector_template: str = None

    def __str__(self) -> str:
        return self.name or self.__class__.__name__

    __repr__ = __str__

```

I don't see the value of typehinting but I appreciate you're a fan of them :)

tests/browser/common_selectors.py
kowalcj0: ah yes of course!
it might be even simpler:
`return getattr(self.value, 'name', None) or self.__class__.__name__`

tests/browser/common_selectors.py
kowalcj0: Ooh yiss I like this.

kowalcj0: re. typehinting
I really like them as they make things very explicit by "self-documenting".
Moreover PyCharm has a great support for it. It already helped me to identify quite a few invalid func invocation before I ran the code.

README.md
kowalcj0: maybe `recreate_db` ?

richtier: :+1:

PR #162, opened 2020-05-11
MVP-450- add image upload feature

depsiatwal:
 - [x] Changelog entry added.

### DIFF
--- a/exportplan/context.py
+++ b/exportplan/context.py
@@ -6,14 +6,9 @@ from exportplan import data
 class ExportPlanDashboardPageContextProvider(AbstractPageContextProvider):
 
     template_name = 'exportplan/export_plan_dashboard_page.html'
-    
+
     @staticmethod
     def get_context_data(request, page):
-        company = None
-        if request.user.is_authenticated and request.user.company:
-            company = request.user.company
-
         return {
             'sections': data.SECTION_TITLES,
-            'company': company
         }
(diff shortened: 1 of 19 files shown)


exportplan/views.py
```diff
@@ -79,3 +84,47 @@ def post(self, request, *args, **kwargs):
...
+        sentry_sdk.set_user(
+            {'hashed_uuid': user.hashed_uuid, 'user_email': user.email}
+        )
+        sentry_sdk.set_extra('api_response', str(api_response.content))
+        sentry_sdk.capture_message('Updating company profile failed')
+
+
+class LogoFormView(BaseFormView):
+    def get_initial(self):
+        return {}
```
eadpearce, line 127: is this supposed to be overriding the method in the base class?

exportplan/context.py
```diff
@@ -1,13 +1,19 @@
...
     @staticmethod
     def get_context_data(request, page):
+        company = None
+        if request.user.is_authenticated and request.user.company:
+            company = request.user.company
+
         return {
-            'sections': data.SECTION_TITLES
+            'sections': data.SECTION_TITLES,
+            'company': company
```
richtier, line 18: no need to pass the company via `get_context_data` - the request object is already exposed to the template, so we can do `request.user.company`

exportplan/views.py
depsiatwal: Yes this is generic base form where for logo we don't need the serialised company

PR #173, opened 2020-05-21
Update signup-wizard-step-start-export-plan.html

logikblok:
NOW!


core/templates/core/signup-wizard-step-start-export-plan.html
```diff
@@ -1,7 +1,7 @@
 {% extends 'core/signup-wizard-step-base.html' %}
 
 {% block above_form %}
-    <h2 class="h-l">Get export plan tailored to your business</h2>
+    <h2 class="h-l">Get export plan tailored to your business NOW!</h2>
```
richtier, line 4: should be lowercase _now!_

PR #187, opened 2020-06-02
Brand and product page + autosave 

eadpearce:

 - [x] Changelog entry added.

Screenshot 2020-06-02 at 17 04 03


exportplan/forms.py
```diff
@@ -15,3 +15,47 @@ class LogoForm(forms.Form):
...
+                'Peak District'
+            }
+        ),
+    )
+    processes = forms.CharField(
+        label='Manufacturing processes',
+        required=False,
+        widget=Textarea(
+            attrs={
+                'placeholder': 'Add some text, for example: The ingredients are ground by hand before being distilled '
```
richtier, line 47: it's a bit hard to differentiate the value from the string. dou you mind doing something  like:

                 'placeholder': (
                     'Add some text, for example: The ingredients are ground by hand before being distilled '
                     'in Cheshire. The alcohol vapour and the flavoursome oils from the botanicals reach our copper '
                      'condenser, where they are immediately cooled'
                )

exportplan/views.py
```diff
@@ -68,6 +76,46 @@ def get_context_data(self, **kwargs):
...
+    def form_valid(self, form):
+        helpers.update_exportplan(
+            sso_session_id=self.request.user.session_id,
+            id=self.export_plan['pk'],
+            data={'brand_product_details': form.cleaned_data}
+        )
+        return super().form_valid(form)
+
+    def get_form_initial(self):
+        return self.export_plan['brand_product_details']
```
richtier, line 92: can use `get_form_initial` instead and not need to call `get_form_kwargs` :sunglasses: 

http://ccbv.co.uk/projects/Django/3.0/django.views.generic.edit/FormView/#get_initial

exportplan/views.py
```diff
@@ -68,6 +76,46 @@ def get_context_data(self, **kwargs):
...
+
+    def get_form_kwargs(self):
+        form_kwargs = super().get_form_kwargs()
+        form_kwargs['initial'] = self.get_form_initial()
+        return form_kwargs
+
+    def get_context_data(self, *args, **kwargs):
+        form_initial = self.get_form_kwargs()['initial']
```
richtier, line 100: `form` is added to the context by the view: http://ccbv.co.uk/projects/Django/3.0/django.views.generic.edit/FormView/#get_context_data

so I thin we could do 

```
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
         form_initial = context['form'].initial
```

richtier: yes ccbv.co.uk is my favorite website :laughing:

exportplan/views.py
```diff
@@ -68,6 +76,46 @@ def get_context_data(self, **kwargs):
...
+
+        field_names = [field[0] for field in self.form_class.base_fields.items()]
```
richtier, line 102: `field_names = [name for name in self.form_class.base_fields]`

or maybe even `field_names = list(self.form_class.base_fields.keys())`

exportplan/views.py
```diff
@@ -68,6 +76,46 @@ def get_context_data(self, **kwargs):
...
+
+        field_labels = [field[1].label for field in self.form_class.base_fields.items()]
```
richtier, line 104:
```
field_labels = [field.label for field in self.form_class.base_fields.values()]
```

can do similar thing for `field_placeholders` too :)

eadpearce: clearly i've been writing too much javascript bc i've forgotten in python you don't have to do everything the hard way LOL...

exportplan/views.py
eadpearce: i need to bookmark this it's much better than django docs

PR #189, opened 2020-06-04
Show demographics data for country

richtier:
 - [x] Changelog entry added.
 -
image


core/helpers.py
```diff
@@ -17,14 +18,70 @@
...
 COMMODITY_SEARCH_URL = urljoin(settings.DIT_HELPDESK_URL, '/search/api/commodity-term/')
+MALE = 'xy'
+FEMALE = 'xx'
 
 
 logger = getLogger(__name__)
 
 
+with open(settings.ROOT_DIR + 'core/fixtures/countries-populations.csv', 'r') as f:
+    country_population_data = f.read()
```
richtier, line 33: previously we would open the file in the request (for every request to the view that uses it). No need to do that as the file contents does not change - so we can load it into memory once and consume it in the view

PR #200, opened 2020-06-24
Fix numbers on curated list page

richtier:
 - [x] Changelog entry added.
 
image


logikblok:
My man

PR #202, opened 2020-06-25
Gp2 72 - Module landing page - overview

lewis-coulson-dit:
Screenshot 2020-06-25 at 14 17 03

 - [x] Changelog entry added.


CHANGELOG.md
```diff
@@ -3,109 +3,112 @@
...
-- MVP-428 - Fix country data component formatting
-- MVP-402 - Fix Localisation
-- MVP-429 - Add missing country thumbnails for food and drink sector
-- MVP-437 - Show product in lesson page nav bar
-- No ticket - v3 ci pipeline manifest.yml file fix
-- MVP-443 - Use non-GREAT stlye 404 and 500 pages
-- MVP-579 - create export plan cms page
-- GP2-67 - fix backbutton upload logo
+
+-   MVP-431 - Fix percentage amounts in country data component
```
eadpearce, line 7: has your editor/linter reformatted these spacings? it makes it hard to see what's actually changed in the diff

react-components/webpack.config.js
```diff
@@ -1,93 +1,94 @@
...
-      // copy assets needed by CSS files as they are not automatically moved to dist foler by React
-      { from: "react-components/assets/stylesheet-assets/", to: "../../core/static/img/" }
-    ]),
-    new RemovePlugin({ after: { include: ["./react-components/dist/img/"] } }),
-  ],
+            {
+                test: /\.(jpg|png|gif|jpeg|woff|woff2|eot|ttf|svg)$/,
+                loader: 'url-loader?limit=10000&name=img/[name].[ext]'
+            }
+        ]
```
eadpearce, line 79: this diff also looks like it's just formatting changes. has anything else changed in this file?

eadpearce:
Please add a screenshot of the UI changes to the PR description

react-components/webpack.config.js
```diff
@@ -53,41 +53,42 @@ module.exports = {
...
-      { from: "./node_modules/great-styles/src/images", to: "images" },
+      { from: './node_modules/great-styles/src/images', to: 'images' },
       // copies the images to core/static only if not present. This avoids
       // the svg files showing up in diff every time a new build occurs
-      { from: "react-components/dist/img/", to: "../../core/static/img/" },
+      { from: 'react-components/dist/img/', to: '../../core/static/img/' },
       // copy assets needed by CSS files as they are not automatically moved to dist foler by React
-      { from: "react-components/assets/stylesheet-assets/", to: "../../core/static/img/" }
+      { from: 'react-components/assets/stylesheet-assets/', to: '../../core/static/img/' },
+      { from: 'design-system/fonts/', to: '../../core/static/img/' }
```
eadpearce, line 90: aha. i can see it now. shouldn't these go into `../../core/static/fonts/` ?

lewis-coulson-dit: I think I tried that but was having issues with incorrect paths. Let me take another look.

core/sass/main.scss
```diff
@@ -7,6 +6,10 @@
 // Config
 $asset-root: '/static/images/';
 
+@import '../../design-system/_style-params.css';
```
eadpearce, line 9: please leave out the `_` and file extension in imports for consistency

react-components/webpack.config.js
eadpearce: you might have to create the directory manually first before it can copy things to it

design-system/_fonts.css
```diff
@@ -0,0 +1,7 @@
+@font-face {
+    font-family: 'FS Lucas';
```
eadpearce, line 2: looking at newer designs we seem to use this font everywhere now - body and headings. you could override the old fonts by setting `$brand-font-heading` and `$brand-font-body` to this. those vars live in https://github.com/uktrade/great-styles

lewis-coulson-dit: These fonts are getting pulled in dynamically from the DSM, not sure if it's likely that the fonts will change or not. If I override those vars now all existing typography which uses the brand font will get updated. Would this be ok?

eadpearce: yep i think that will be fine. we weren't using the correct fonts anyway lol

lewis-coulson-dit: Just thinking about this and not sure if this is going to work well with the DSM. Right now I have added a new file called `_mixins.scss` to the design system. I expose different mixins that make use of the CSS variables generated from the DSM. If I move the fonts into the great-styles project then that will become a hard dependency in the `_mixins.scss` file. This could be a bit confusing since we will be using variables generated from the DSM and also sass vars from great-styles project. Does that make sense?

react-components/webpack.config.js
lewis-coulson-dit: cheers that worked as well as updating the webpack config to add the correct path for fonts.

core/sass/main.scss
lewis-coulson-dit: Cleaned this up now

design-system/_fonts.css
lewis-coulson-dit: Ignore above, I am working on the next module landing ticket and I will need to update some styles in the great-styles project anyway. I'll update the font variables at the same time as other changes for ticket GP2-73.

PR #203, opened 2020-06-26
save hs_codes on Company in directory_api

WillTaylor22:
- [x] Changelog entry added.


react-components/src/views/ProductsModal/Container.jsx
```diff
@@ -35,7 +35,7 @@ export function Container(props){
   }
 
   function buildNextUrl() {
-    const reducer = (accumulator, product) => `${accumulator}&product=${product.label}&`
+    const reducer = (accumulator, product) => `${accumulator}&product=${product.label}&hs_codes=${product.value}`
```
richtier, line 38: needs to end with `&` as the return value is passed back into the reducer in the form of `accumulator`

core/middleware.py
```diff
@@ -52,9 +52,13 @@ def should_set_product_expertise(self, request):
...
         if self.should_set_product_expertise(request):
             products = request.GET.getlist('product')
+            hs_codes = request.GET.getlist('hs_codes')
             helpers.update_company_profile(
                 sso_session_id=request.user.session_id,
-                data={'expertise_products_services': {'other': products}}
+                data={
+                    'expertise_products_services': {'other': products},
+                    'hs_codes': hs_codes
+                }
```
richtier, line 61: please add/update test

PR #208, opened 2020-07-01
Update README.md

lewis-coulson-dit:
Adds some updates to readme relevant for new joiners


README.md
```diff
@@ -16,7 +16,9 @@
     $ git clone https://github.com/uktrade/great-cms
     $ cd great-cms
     $ [create and activate virtual environment]
+    $ python -m pip install pip-tools
```
eadpearce, line 19: pip-tools is already included in requirements_test so it should be installed when running `make install_requirements`

lewis-coulson-dit: ok cool will remove

PR #221, opened 2020-07-08
Objectives list

eadpearce:

 - [x] Changelog entry added.

Screenshot 2020-07-08 at 17 35 18

Screenshot 2020-07-08 at 17 57 04


exportplan/api.py
```diff
@@ -93,3 +95,56 @@ def get(self, request):
...
+            helpers.update_exportplan(
+                sso_session_id=self.request.user.session_id,
+                id=export_plan['pk'],
+                data=serializer.validated_data
+            )
+            return Response(serializer.validated_data)
+
+
+class ObjectivesCreateAPIView(generics.GenericAPIView):
+    serializer_class = serializers.NewObjectiveSerializer
```
depsiatwal, line 118: why NewObjectivesSerializer ?
can we just call it ObjectivesSerializer

exportplan/serializers.py
```diff
@@ -27,5 +27,34 @@ def validate_target_markets(self, values):
...
+        return value
+
+    def validate_end_date(self, value):
+        if value == '':
+            return None
+        return value
+
+
+class NewObjectiveSerializer(ObjectiveSerializer):
+    pk = serializers.IntegerField(required=False)
```
depsiatwal, line 52: may be remove and user pkSerializer

exportplan/api.py
eadpearce: it's for creating a new objective so the only difference between it and ObjectiveSerializer is that the pk is not required

depsiatwal: what about making it optional and have a validate method ?

exportplan/serializers.py
eadpearce: needs to have all the fields that ObjectiveSerializer has as well. only difference is pk is not required here

exportplan/api.py
depsiatwal: ahh didn't realise it inherits for objectiveserializer

README.md
```diff
@@ -57,10 +57,10 @@ Secrets such as API keys and environment specific configurations are placed in `
...
 | make css                      | Compile scss to css |
 | make secrets                  | Create your secret env var file |
-| make recreate                 | Runs following command in one go |
-|                               | **make database**: Drop and recrete the database |
-|                               | **make boostrap_great**: Create required database records so the CMS works |
-|                               | **make create_tours** |
+| make database                 | Drop and recreate the database |
+| make bootstrap_great          | Create required database records so the CMS works |
+| make create_tours             | Create required Page tours |
+| make recreate                 | Runs: database, bootstrap_great, create_pages, create_tours in one go |
```
webbyfox, line 63: This was recently updated on Readme. As some of these commands aer not standalone `make` command hence grouped them.

eadpearce: oh i think that must have been a mistake when rebasing. i can fix that

PR #224, opened 2020-07-10
update content of the brand and product page, add missing data, updat…

dahfool:
…e folder structure for fields and fieldWithExample

 - [ ] Changelog entry added.


exportplan/forms.py
```diff
@@ -41,19 +41,25 @@ class CountryDemographicsForm(forms.Form):
...
+                'Dove Gin was founded in 2012 when Simon Dove started to distill gin in his garage in '
+                'Shrewsbury. Simon came across a book of gin recipes on a visit to The Gin Museum of London. '
+                'This inspired him to recreate gin distilled in the midlands 200 hundred years ago, with a '
+                'modern twist.'
             )}
         ),
     )
     location = forms.CharField(
-        label='Location',
+        label="Where we\'re based",
```
eadpearce, line 62: no need to escape that quote when the whole string is in double quotes

dahfool: Thanks. I have udpated.

PR #230, opened 2020-07-15
create target market research page

dahfool:

 - [ ] Changelog entry added.


exportplan/views.py
```diff
@@ -140,6 +140,23 @@ def form_valid(self, form):
         return super().form_valid(form)
 
 
+class ExportPlanTargetMarketResearchView(FormContextMixin, ExportPlanSectionView, FormView):
+
+    def get_initial(self):
+        return self.export_plan['brand_product_details']
```
depsiatwal, line 146: This can be changed to 
return self.export_plan['target_markets_research'

exportplan/views.py
```diff
@@ -140,6 +140,23 @@ def form_valid(self, form):
...
+
+    form_class = forms.ExportPlanTargetMarketResearchForm
+    success_url = reverse_lazy('exportplan:target-market-research')
+
+    def form_valid(self, form):
+        helpers.update_exportplan(
+            sso_session_id=self.request.user.session_id,
+            id=self.export_plan['pk'],
+            data={'target_market_research': form.cleaned_data}
```
depsiatwal, line 155: data={'target_markets_research': form.cleaned_data}

exportplan/views.py
dahfool: @depsiatwal This change will break my local as i don't have the target_markets_research field in the db.

PR #232, opened 2020-07-20
Gp2 192 - Marketing approach - Structure of the page

lewis-coulson-dit:
New Marketing approach page structure


lewis-coulson-dit:
> I'm just wondering why there is a massive reduction in the compiled react components .js file?
> 84K deletions sounds wrong somehow.
> I took a quick look and it seems like the line-feeds have been removed.

@bobmeredith This is because running `npm run watch` and `npm run build` produce different outputs to the compiled CSS and Javascript. My guess is that the `npm run watch` output was pushed into `develop` hence why so many deletions on my branch.

PR #233, opened 2020-07-20
Gp2 207 welcome back

bobmeredith:

 - [x] Changelog entry added.
 - [x] Upgraded any vulnerable dependencies.
 - [x] Requirements have been compiled.


core/templates/core/dashboard.html
```diff
@@ -1,9 +1,10 @@
 {% extends 'core/base.html' %}
+{% load to_json %}
```
webbyfox, line 2: Is this been used?

sso/models.py
```diff
@@ -39,3 +39,19 @@ def save(self, *args, **kwargs):
     @cached_property
     def export_plan(self):
         return get_exportplan(self.session_id)
+
+    @cached_property
+    def user_profile(self):
+        return helpers.get_user_profile(self.session_id)
```
webbyfox, line 45: wondering why they have used `session_id` as it implies dynamic for each session. Actual it is `sso_use_id`

tests/unit/sso/test_models.py
```diff
@@ -1,13 +1,44 @@
...
+    assert user.update_user_profile({'firstname': 'Jim'}) == ok_result
+
+
+@mock.patch.object(sso_api_client.user, 'get_user_page_views')
+def test_has_visited_page(mock_get_user_page_views):
+    user = get_user()
+    page_views_response = {'result': 'ok', 'page_views': {'mypage': 1}}
+    mock_get_user_page_views.return_value = create_response(page_views_response)
+    assert user.get_page_views() == page_views_response
+    assert user.has_visited_page(page='mypage') == {'mypage': 1}
```
webbyfox, line 44: Do we need a negative test? For example,  user not visited page and check `has_visited_page`

core/templates/core/dashboard.html
bobmeredith: Fair point.
No it's not ATM, but it's a useful thing to show context while building up the page.
I'll take it out of this PR.

sso/models.py
bobmeredith: Yes, It is the sso_id, but the field in  BusinessSSOUser is called 'session_id'. It's static and unique to a user as it's the key used to retrieve the user profile from directory_sso.
I don't fancy doing a renaming exercise in this PR.

tests/unit/sso/test_models.py
bobmeredith: Could do.  The tests in directory_sso do negative tests, but there's no harm in one here too. 
Will add one - as a helper test though.
Essentially, the code in great-cms can either get all page views for this user/service and check for pages in the result, or filter the return to look at just one page. In the future, we may want to put the whole list in redis to speed things up.

webbyfox: Yeah, I dont want as well but just observation.

sso/models.py
webbyfox: Yeah, I dont want as well but just observation.

webbyfox:
LGTM, Just update the branch and good to merge

PR #234, opened 2020-07-20
update fieldwithExample component to handle different variation

dahfool:

 - [x] Changelog entry added.


dahfool:
> we'll need tests for the views else code cov will fail, let me know if you need a hand :-)

Yes, i will need some help.

PR #240, opened 2020-07-23
git add a new field for the market research page

dahfool:

 - [ ] Changelog entry added.


dahfool:
> I suspect tests will need to be updated
> and also Change long update please

Change log and FE test updated. Do we need any BE tests for the changes i have made.

dahfool:
> also maybe add your field in this test
> 
> https://github.com/uktrade/great-cms/blob/develop/tests/unit/exportplan/test_serializers.py#L4

Unit test added

PR #244, opened 2020-07-27
Gp2 156 - Learning onboarding - welcome message

lewis-coulson-dit:


react-components/src/views/LoginModal/Step1.jsx
```diff
@@ -42,20 +46,21 @@ export default function Step1(props){
...
         />
-        <p><a href={Services.config.passwordResetUrl}>Forgotten password?</a></p>
+        <p>
+          <a href={Services.config.passwordResetUrl}>Forgotten password?</a>
+        </p>
       </form>
 
-      <div className='great-mvp-vertical-separator'>
-        <hr/>
+      <div className="great-mvp-vertical-separator">
```
dahfool, line 54: @lewis-coulson-dit are we using single or double quotes ? `great-mvp-vertical-separator` might need to be renamed.

lewis-coulson-dit: Seems to be double quotes on my prettier setup at least. Looks mixed throughout code base right now.

PR #246, opened 2020-07-27
GP2-159 - Learning onboarding - modules tooltip

lewis-coulson-dit:


react-components/src/views/Tour/Tour.jsx
```diff
@@ -5,33 +5,46 @@ import ReactTour from 'reactour'
...
+      isOpen={props.isOpen}
+      onRequestClose={props.handleClose}
+      showNumber={false}
+      lastStepNextButton={
+        <button className="button button--primary p-v-xxs p-h-xs" id="page-tour-start-now">
+          Start now
+        </button>
+      }
+      nextButton={
+        <button className="button button--primary p-v-xxs p-h-xs" id="page-tour-next-step">
```
bobmeredith, line 41: Just a question - can you use these styles like `button--primary` from DSM/great styles yet? I don't see them in the latest published great-styles.
Unless I'm missing something.

dahfool:
@lewis-coulson-dit I know this component was not created by you, but should we fix es-lint issues add unit tests ..... when we update components.

react-components/src/views/Tour/Tour.jsx
lewis-coulson-dit: Yup for the elements I have added. What version of great-styles are you using?

lewis-coulson-dit:
@dahfool Yes we probably should be. I have one more ticket related to this work I'm about to start. I will update the unit tests and fix and listing on that ticket.

react-components/src/views/Tour/Tour.jsx
bobmeredith: I've only got 0.1.0 but great csm package is "great-styles": "uktrade/great-styles#v0.2.0",
Something wrong there - I'll try installing again.

PR #253, opened 2020-07-29
Gp2 210 personalisation bar restyle

bobmeredith:

 - [Y] Changelog entry added.
 - [N] I have updated security dependencies


learn/templates/learn/detail_page.html
```diff
@@ -32,84 +32,28 @@ <h3 class="h-s">{{ topic.title }}</h3>
...
-                            {% endfor %}
-                        {% else %}
-                            <span class="great-mvp-pill-button">add countries</span>
+                {% include "learn/includes/personalisation_bar.html" with profile=request.user.company %}
+            </div>
+            <div class="grid p-t-l p-f-l">
+                <div class="c-2-3">
+                    <h1 class="h-l" id="lesson-title">{{ self.title }}</h1>
+                    {% include 'core/personalised_body.html' %}
+                    <div id="countries-chooser" class="great-mvp-full-width-background"></div>
```
lewis-coulson-dit, line 41: I think we want to start removing any class names with an `mvp` in them now so change `great-mvp-full-width-background` to `great-full-width-background`

learn/templates/learn/includes/personalisation_bar.html
```diff
@@ -0,0 +1,52 @@
+
+{% load wagtailimages_tags %}
+{% load routablepageurl from wagtailroutablepage_tags %}
+
+<nav class="p-t-xs p-h-xs p-b-xxs great-mvp-full-width-background" id="market-select-navbar">
```
lewis-coulson-dit, line 5: Same here

learn/templates/learn/detail_page.html
bobmeredith: Fair enough. However, I don't see the point of leaving styles like this in great-cms. This looks like a candidate for a helper class.  I can't find one that quite does the job but I could slightly modify one.

bobmeredith: Added class w-full-viewport to great styles and deleted the old mvp style.
They'll be a lot more of this to do!

PR #255, opened 2020-07-30
GP2 161 - Learning onboarding - turn off onboarding feature

lewis-coulson-dit:


react-components/src/views/Tour/Modal.jsx
```diff
@@ -58,11 +58,19 @@ export default function Modal(props) {
           </a>
         </div>
       </div>
+      <div className="great-tour-modal__footer">
+        <div className="great-checkbox">
+          <input type="checkbox" id="great-tour-disable" onChange={() => props.handleDisable()} />
+          <label htmlFor="great-tour-disable">Don’t show me this again</label>
+        </div>
+      </div>
```
bobmeredith, line 66: I hate to mention it but I notice the tour component is tested in the pytest_browser suite on great-cms.  Should the disable be tested as part of this?

dahfool:
@lewis-coulson-dit @bobmeredith we need to add storybook to great-cms so that it's easier to see which components exists and reuse.

PR #257, opened 2020-08-03
hook-up market-search target-audience

depsiatwal:
 - [x] Changelog entry added.


webbyfox:
Looks good. might need to add a unit test to cover `raise` statement in serializer make Codecov happy

PR #259, opened 2020-08-03
Gp2 194 route to market component

dahfool:

 - [ ] Changelog entry added.

### DIFF
--- a/react-components/src/components/RouteToMarket/RouteToMarket.jsx
+++ b/react-components/src/components/RouteToMarket/RouteToMarket.jsx
@@ -1,81 +1,45 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import ReactDOM from 'react-dom'
+
+import { RouteToMarketSection } from './RouteToMarketSection'
+import { addItemToList } from '../../Helpers'
 import './RouteToMarket.scss'
 
-function renderTable({ route, promote, why }, index, updateTable) {
+const RouteToMarket = (props) => {
+
+  const [rows, setRows] = useState([])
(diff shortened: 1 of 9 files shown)


react-components/src/components/RouteToMarket/RouteToMarket.jsx
```diff
@@ -1,81 +1,45 @@
...
       ]
-    })
+    )
   }
 
-  updateTable = (index, key, value) => {
-    const newState = this.state.tables
+  useEffect(() => {
+    if (data.length > 0) console.log('POST DATA HERE',data)
+  }, [data])
```
bobmeredith, line 24: I guess this useEffect can go now.

dahfool:
> LGTM! nice work, the only thing I noticed is inconsistent spacing and line breaks. Maybe run prettier?

Thanks, i am using the one in the repo. Do we need to update ?

{
  "singleQuote": true,
  "printWidth": 120,
  "arrowParens": "always",
  "semi": false,
  "tabWidth": 2,
  "useTabs": false
}

react-components/src/components/RouteToMarket/RouteToMarket.jsx
dahfool: > I guess this useEffect can go now.

The `useEffect` check whether the data has been updated and do an ajax call to the backend. The `console.log` is for Gurdeep, so that he knows where to add the function to update the BE.

bobmeredith: Ah OK.  Makes sense.

PR #263, opened 2020-08-04
 GP2-195-marketing-approach-ui

dahfool:

 - [x] Changelog entry added.


dahfool:
> Can you make the PR name reflect something similar to the branch name ?.
> 
> Also the form appear on a page
> 
> 1. Create a view similar to
>    https://github.com/uktrade/great-cms/blob/develop/exportplan/views.py#L141
> 2. Also check the template above and do same to the template you wish to add too.
> 3. form and view will need tests similar have a look under unittests/exportplan
>    test_viewss.py and test_forms.py

There is already a `ExportPlanMarketingApproachView`

PR #272, opened 2020-08-07
create adaptation for target market page

dahfool:

 - [x] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 - GP2-170 - business performance
 GP2-315 - route to market backend integration
 ### Implemented enhancements
-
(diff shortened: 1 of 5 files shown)


dahfool:
> will need to add tests :-)

dahfool:
> will need to add tests :-)

For the view ?

depsiatwal:
yeah any python code will need tests else codecov fails
you can see under unit/tests. test_views classes

depsiatwal:
it might already be covered as codecov hasn't detected a drop actaully

dahfool:
> it might already be covered as codecov hasn't detected a drop actaully

🤷‍♂️

PR #273, opened 2020-08-09
Feature/aa 129 beta env great with ga360

agiamas:

 - [X] Change has a jira ticket that has the correct status.
 - [X] Changelog entry added.
 - [X] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [X] (if updating requirements) Requirements have been compiled.


tests/unit/core/test_views.py
```diff
@@ -14,8 +14,12 @@
 from tests.helpers import create_response
 from tests.unit.core.factories import CuratedListPageFactory, DetailPageFactory, ListPageFactory
 from tests.unit.learn.factories import LessonPageFactory
+from django.http.cookie import SimpleCookie
 from tests.unit.domestic.factories import DomesticDashboardFactory
 
+BETA_AUTH_TOKEN_PAST = 'gAAAAABfCpH53lJcM0TiiXTqD7X18yRoZHOjy-rbSogRxB0v011FMb6rCkMeizffou-z80D9DPL1PWRA7sn9NBrUS' \
```
depsiatwal, line 20: should this not be in setting / config ?

agiamas: no, this is just for testing with an invalid token (b/c it's like expired a month ago)

PR #275, opened 2020-08-10
GP2-257 - Step by step

lewis-coulson-dit:


learn/templates/learn/detail_page.html
```diff
@@ -62,11 +65,59 @@ <h5>Next</h5>
...
+
+            <hr class="hr--thick" />
+
+            <div class="grid p-t-l p-f-l">
+                <div class="c-1-4">
+                    <h4 class="h-s">&nbsp;</h4>
+                </div>
+                <div class="c-1-2">
+                    <ul class="step-by-step">
+                        <li class="media-block m-b-m">
```
bobmeredith, line 81: I'm a bit concerned about putting hard-coded stuff in templates like this.
This page should be completely generic with no hard-coded content at all I thought. The body is a rich text area already.
So, i would have thought we should maybe be chopping the structure of lesson pages into bits, and building the template to glue them together - in which case, any hard-coded stuff should be in the view - pending putting into wagtail.
Same goes for the lorem-ipsum before this.

PR #279, opened 2020-08-10
GP2-317-order-of-landing-pages

depsiatwal:
 - [x] Changelog entry added.


exportplan/views.py
```diff
@@ -152,7 +149,8 @@ def get_context_data(self, *args, **kwargs):
         return context
 
 
-class ExportPlanTargetMarketsView(ExportPlanSectionView):
+class ExportPlanTargetMarketsView(TemplateView):
+    # This view has been taken our of scope leaving it here for now as it may get reintroduceded
```
webbyfox, line 153: maybe typo? `out of scope` and `reintroduced`

PR #281, opened 2020-08-11
GP2-195-FE-BE-integration

dahfool:

 - [x] Changelog entry added.


react-components/dist/ditMVP.js
```diff
@@ -85173,6 +85177,7 @@ var createRouteToMarket = function createRouteToMarket(_ref) {
   var element = _ref.element,
       params = _objectWithoutProperties(_ref, ["element"]);
 
+  debugger;
```
lewis-coulson-dit, line 85180: guessing you want this removed?

dahfool: 👍  good catch, i never check the compiled js.

PR #283, opened 2020-08-12
Added functionality to complete lesson

webbyfox:

 - [x] Changelog entry added.
 - [x] Requirements have been compiled.


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Changelog
+- AA-129 - Beta environment for GREAT.gov
```
depsiatwal, line 2: This need to go under enhancements

sso/helpers.py
```diff
@@ -135,6 +136,39 @@ def get_user_page_views(sso_session_id, page=None):
...
+def get_lesson_completed(sso_session_id, lesson=None):
+    response = sso_api_client.user.get_user_lesson_completed(sso_session_id, SERVICE_NAME, lesson)
+    if response.status_code in [400, 404]:
+        raise APIException(detail=response.json(), code=response.status_code)
+    response.raise_for_status()
+    return response.json()
+
+
+def delete_lesson_completed(sso_session_id, lesson=None):
+    response = sso_api_client.user.delete_user_lesson_completed(sso_session_id, SERVICE_NAME, lesson)
```
bobmeredith, line 165: Question - so, if no lesson is passed to delete - all lesson completes for this user/service are deleted right?

webbyfox: No, it will give bad request as now only one lesson can be deleted.

depsiatwal: then why provide a default 
same applies to all the other methods ?

PR #284, opened 2020-08-12
gp2-369-CHEG-tailored-link

depsiatwal:
 - [x] Changelog entry added.


exportplan/urls.py
```diff
@@ -13,15 +13,10 @@
...
-        name='target-markets'
-    ),
-    path(
-        'section/adaptation-for-target-market/',
+        'section/adaptation-for-your-target-market/',
         login_required(views.ExportPlanAdaptationForTargetMarketView.as_view(), login_url=LOGIN_URL),
-        {'slug': 'adaptation-for-target-market'},
-        name='adaptation-for-target-market'
+        {'slug': 'adaptation-for-your-target-market'},
+        name='adaptation-for-your-target-market'
```
webbyfox, line 19: Is `adaptation-for-target-market` namespace been used anywhere in codebase to resolve URL? If not then this is fine.

depsiatwal: It's create in and reference in data.SECTION_TITLES which has also been changed so change is good :-)

PR #287, opened 2020-08-13
Gp2 365 target market service page

rafa-garcia:
 - [x] Changelog entry added.


core/models.py
```diff
@@ -239,6 +239,38 @@ class LandingPage(CMSGenericPage):
     ]
 
 
+class ExtendedLandingPage(CMSGenericPage):
```
bobmeredith, line 242: So I just tried adding the generic template to the original landing page, and AFAICT, it works just fine.
The landing page is already in the core app - so it doesn't look like we need this new page at all - merely add the template and include it in the list of templates available in the landing page model and bingo.
`    template_choices = (
        ('learn/landing_page.html', 'Learn'),
        ('core/extended_landing_page.html', 'Generic'),
    )`
The only thing that's slightly naff is the name of the page - it should possibly be more like 'generic page' than 'Landing page' but heh ho!

bobmeredith:
@rafa-garcia Could you leave a comment when you've reviewed this - as you probably can't review yourself.

core/models.py
```diff
@@ -229,13 +230,26 @@ class LandingPage(CMSGenericPage):
         related_name='+'
     )
 
+    body = StreamField([
+        ('section', core_blocks.SectionBlock()),
+        ('title', core_blocks.TitleBlock()),
+        ('text', blocks.RichTextBlock(icon='openquote', helptext='Add a textblock')),
+        ('image', core_blocks.ImageBlock()),
+    ], null=True, blank=True)
```
webbyfox, line 238: This will make body field optional. Is this expected?

rafa-garcia: It is expected, this image(s) is not to be mistaken for the page's main image

rafa-garcia:
Looks great @bobmeredith 👍

core/models.py
bobmeredith: Yes, optional is good - this is the landing page so it could just have the title, description and a button.

PR #288, opened 2020-08-13
No-ticket_fix_for_wagtail_editor

bobmeredith:

 - [N/A] Change has a jira ticket that has the correct status.
 - [N/A] Changelog entry added.

Tiny fix to stop error in wagtail editor

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -174,10 +174,10 @@ class CheckGATags(MiddlewareMixin):
         if getattr(response, 'skip_ga360', False):
             return response
 
-        if not hasattr(response, 'context_data'):
+        if not hasattr(response, 'context_data') or not response.context_data:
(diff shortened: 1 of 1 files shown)


core/middleware.py
```diff
@@ -177,8 +177,7 @@ def process_response(self, _, response):
         if not hasattr(response, 'context_data'):
             logger.error('No context data found')
             return response
-        context_data = response.context_data
-
+        context_data = response.context_data or {}
```
webbyfox, line 180: I think if `context_data` is None then we should return response so adding once more check on if statement would be ideal? what do you think?

bobmeredith: You're quite right.
I threw this in a quick fix so I could edit a page. 
 It could be handled better in the block above
One sec ...

PR #291, opened 2020-08-14
Gp2 36 - Mark as complete

lewis-coulson-dit:

 - [x] Changelog entry added.


react-components/src/components/MarkLessonAsComplete/MarkLessonAsComplete.jsx
```diff
@@ -0,0 +1,52 @@
...
+        .then((response) => response.json())
+        .then(({ lesson_completed }) => {
+          if (lesson_completed.length >= 1) {
+            setIsComplete(true)
+          }
+        })
+        .catch((e) => console.log('error: ', e))
+    } else {
+      Services[isComplete ? 'setLessonComplete' : 'setLessonIncomplete'](endpoint).catch((e) =>
+        console.log('error: ', e)
```
dahfool, line 22: Is the `console.log` intentional ?

lewis-coulson-dit: good spot, have replaced with empty arrows functions.

react-components/src/components/MarkLessonAsComplete/MarkLessonAsComplete.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { useState, useEffect } from 'react'
+import ReactDOM from 'react-dom'
+
+import Services from '../../Services'
+
+const MarkLessonAsComplete = ({ endpoint }) => {
+  const [isComplete, setIsComplete] = useState(undefined)
```
dahfool, line 7: is isComplete a boolean ? If so you can initialised by setting `useState(false)`

lewis-coulson-dit: Yea I set it to undefined initially so I can do the initial fetch in the `useEffect` hook

dahfool: Does it have 3 states, `undefined`, 'false' and 'true'

dahfool: 👍

lewis-coulson-dit: I can add a different var like `hasLoaded` I guess, that just seemed like a nice way of differentiating from the initial load state and the toggle states.

PR #297, opened 2020-08-17
Gp2 195 route to market refactor

dahfool:

 - [ ] Changelog entry added.


react-components/src/components/ObjectivesList.jsx
```diff
@@ -199,6 +199,7 @@ class ObjectivesList extends React.Component {
 }
 
 function createObjectivesList({ element, ...params }) {
+  debugger
```
bobmeredith, line 202: debugger statement left in

dahfool: Removed, thanks @bobmeredith

PR #298, opened 2020-08-17
Gp2 257 step by step

webbyfox:
 - [x] Changelog entry added.


webbyfox:
Yes, It's just adding a field to a model which should cover in wagtail testing itself.

PR #306, opened 2020-08-19
GP2-415_fix_link_button_paths

bobmeredith:

 - [Y] Changelog entry added.

### DIFF
--- a/core/blocks.py
+++ b/core/blocks.py
@@ -119,11 +119,12 @@ class SidebarLinkBlock(blocks.StructBlock):
     def render(self, value, context={}):
         try:
             internal_link = value['link']['internal_link']
-            page = models.DetailPage.objects.get(id=internal_link.id)
-            value['target_lede'] = page.get_parent() and page.get_parent().title
-            value['target_title'] = page.title
-            value['read_time'] = getattr(page, 'estimated_read_duration')
-        except (ObjectDoesNotExist, KeyError, TypeError):
+            value['target_lede'] = internal_link.get_parent() and internal_link.get_parent().title
+            value['target_title'] = internal_link.title
+            # If it's a detail page, get the get the read duration
+            detailPage = (models.DetailPage.objects.get(id=internal_link.id))
+            value['read_time'] = getattr(detailPage, 'estimated_read_duration')
+        except (ObjectDoesNotExist, KeyError, TypeError, AttributeError):
             pass
         return super().render(value, context=context)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed bugs
 
+- GP2-415 - internal button links
 - GP2-401 - lesson page rich text styling
 - No ticket - Remove css vars and mixins
 - GP2-257 - Added Step fields to Detail Model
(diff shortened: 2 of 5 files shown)


core/blocks.py
```diff
@@ -119,11 +119,12 @@ class SidebarLinkBlock(blocks.StructBlock):
...
             internal_link = value['link']['internal_link']
-            page = models.DetailPage.objects.get(id=internal_link.id)
-            value['target_lede'] = page.get_parent() and page.get_parent().title
-            value['target_title'] = page.title
-            value['read_time'] = getattr(page, 'estimated_read_duration')
-        except (ObjectDoesNotExist, KeyError, TypeError):
+            value['target_lede'] = internal_link.get_parent() and internal_link.get_parent().title
+            value['target_title'] = internal_link.title
+            # If it's a detail page, get the get the read duration
+            detail_page = (models.DetailPage.objects.get(id=internal_link.id))
```
webbyfox, line 125: shall add a condition to check if it `DetailPage` then only get `detail_page`?

core/blocks.py
```diff
@@ -119,11 +119,12 @@ class SidebarLinkBlock(blocks.StructBlock):
...
+            value['read_time'] = getattr(detail_page, 'estimated_read_duration')
+        except (ObjectDoesNotExist, KeyError, TypeError, AttributeError):
```
webbyfox, line 127: If `DetailPage` then you might not required `AttributeError` exception as `DetailPage` should have `read_time`?

tests/unit/core/test_blocks.py
```diff
@@ -90,15 +90,15 @@ def test_learning_link_component(domestic_homepage):
...
     # Override the title and lede
     result_override = link_block.render(
         value={
             'title_override': override_title,
             'lede_override': override_lede,
-            'link': {'internal_link': target_page}
+            'link': {'internal_link': target_page, 'external_link': ''}
         },
         context={})
     assert override_title in result_override
```
webbyfox, line 104: we should assert the link as well or separate test for link_block?

core/blocks.py
bobmeredith: This is true
will update

tests/unit/core/test_blocks.py
bobmeredith: Indeed - so I added a ListPage as well and the DetailPage and am checking links and read duration now.

core/blocks.py
webbyfox: `AttributeError` still in list?

bobmeredith: AttributeError gets thrown if the link is not populated `value['link']['internal_link']` i.e. None has no attribute `internal_link`
There is nothing more to do in this case.  I could maybe make the link mandatory but that's maybe inconvenient for content designers, and anyway, the code should be defensive, given that a streamfield is, by definition, not fully constrained.

webbyfox: Ideally, you check variable before you accessing attributes. so you don't cause unnecessary attribute error and catch them later, which is extra parsing. However, I will mark this as resolved.

webbyfox (approved):
Test coverage needs to improve although PR looks good

PR #311, opened 2020-08-24
GP2-424-remove-menu-bar-links-logged-out

depsiatwal:
 - [x] Changelog entry added.


tests/browser/test_home_page.py
```diff
@@ -82,7 +79,6 @@ def should_see_expected_error_messages(
 
 def test_anonymous_user_should_not_see_header_elements_for_authenticated_users(browser, visit_home_page):
     should_not_see_errors(browser)
-    should_see_all_expected_page_sections(browser, [HeaderCommon, HeaderSignUp])
     should_not_see_any_element(browser, HeaderSignedIn)
```
bobmeredith, line 82: Shouldn't the user still see the sign-up link in the header?
There is a sign-up link in the login modal too I guess.

bobmeredith (approved):
Looks OK, but I'm not sure if the sign-up link should still be available in the header when logged out. In fact - should sign-up be there when the user is logged in?

tests/browser/test_home_page.py
depsiatwal: This change is about hiding the menu bar for logged-out user hense update in tests

PR #313, opened 2020-08-24
GP2-438 - lesson recap

lewis-coulson-dit:

 - [ x] Change has a jira ticket that has the correct status.
 - [ x] Changelog entry added.


bobmeredith (approved):
LGTM

PR #317, opened 2020-08-25
GP2-438 - lesson fictional example

lewis-coulson-dit:

 - [x] Changelog entry added.


core/sass/learn/_base.scss
```diff
@@ -338,6 +338,13 @@ $desktop: 1024px;
     img, video {
       border-radius: 5px;
       max-width: 100%;
+      width: 100%;
+    }
+  }
+
+  .fictional-example {
+    h2, h3, h4 {
+      margin-bottom: 12px;
```
bobmeredith, line 347: Just a thought - could you do this with helper classes in the markup instead?  I know you can't get to 12px though.

bobmeredith (approved):

core/sass/learn/_base.scss
lewis-coulson-dit: Yes that would be lovely... unfortunately have to have these overrides because wagtail spits out just HTML elements without classes from the rich text editor :(

PR #333, opened 2020-09-02
Gp2 208 product search modal

bobmeredith:

 - [x] Changelog entry added.
 - [x] (if adding env vars) Added any new environment variable to vault.


config/env/secrets-template
```diff
@@ -24,4 +24,6 @@ DIRECTORY_FORMS_API_SENDER_ID=debug#DO NOT ADD SECRETS TO THIS FILE
 AIRTABLE_API_KEY=debug#DO NOT ADD SECRETS TO THIS FILE
 #DO NOT ADD SECRETS TO THIS FILE
 DIT_HELPDESK_URL=http://debug.com#DO NOT ADD SECRETS TO THIS FILE
+#DO NOT ADD SECRETS TO THIS FILE
+3CE_COMMODITY_SEARCH_TOKEN=debug#DO NOT ADD SECRETS TO THIS FILE
```
depsiatwal, line 28: probably not ideal to name variables with a number

config/settings.py
```diff
@@ -431,6 +431,11 @@
     'MADB_URL', 'https://www.check-duties-customs-exporting-goods.service.gov.uk'
 )
 
+# 3CE commodity classification
+COMMODITY_SEARCH_URL = 'http://info.dev.3ceonline.com/ccce/apis/classify/v1/interactive/classify-start'
+COMMODITY_SEARCH_REFINE_URL = 'http://info.dev.3ceonline.com/ccce/apis/classify/v1/interactive/classify-continue'
```
depsiatwal, line 436: Should this not be set as an env variable ?

lewis-coulson-dit (approved):

config/settings.py
bobmeredith: Good question. I thought about this, but reached the conclusion that hard-coding is better on the basis that 3ce don't provide distinct develop/live endpoints (rather they use the bearer token to determine access) and thus the urls would be the same in all environments.  If the urls were fed in from vault, it adds more moving parts, and we would need to set the same URLs the same in all environments.
I only just noticed the 'dev' in the URL, but these URLs are printed in their API documentation .pdf so there's really only these.
Also, took the lead from the MAPDB url above.

config/env/secrets-template
bobmeredith: Good point.  TBH, it didn't cross my mind.  How important do you think this is to change?

config/settings.py
webbyfox: Bob, still a good idea to keep this in env to minimise further development in future and keeping the same practise and align with other APIs, what do you think?

webbyfox (commented):
Apart from the above comments its looks really good

react-components/src/components/ProductFinder/CountryFinder.jsx
```diff
@@ -0,0 +1,115 @@
...
+    modalContent.style.maxHeight='700px'
+  }
+
+  function saveCountry(country) {
+    setSelectedCountry(country);
+    let result = Services.updateExportPlan({export_countries:[country]}).then((result) => {
+      closeModal();
+      // window.location.reload(true);   # do we need to reload page
+    }).catch((result) => {
+      debugger;
```
dahfool, line 48: Please remove debugger and code commented out.

config/settings.py
depsiatwal: Agree with Rizwan about consistently. Also if URLS changed then we we would require a release 
may be low risk for a single URL but we all external sources at one point one of them change domains/urls. 
We can get around setting in all environments by setting a default in settings and only add to Vault to override.

config/env/secrets-template
depsiatwal: I think either we allow this or we don't 
https://www.w3schools.com/python/python_variables.asp says it's bad :-) 
surprised flake8 didn't pick it up

depsiatwal (approved):
LGTM

config/env/secrets-template
bobmeredith: Yes, have changed to CCCE instead.

config/settings.py
bobmeredith: NP, will set as env variables with a default in settings. We won't need to define in Vault unless the urls change for some reason.

react-components/src/components/ProductFinder/CountryFinder.jsx
bobmeredith: Will do.  Nothing to put in api catches yet though - I'll add  TODOs and do whatever the design says when there's a design.

PR #338, opened 2020-09-03
GP2-382_remove_None_on_no_firstname

bobmeredith:

 - [Y] Changelog entry added.


webbyfox (approved):
LGTM

PR #339, opened 2020-09-03
GP2-420-display-lesson-content

depsiatwal:
 - [x] Changelog entry added.


learn/templates/learn/curated_list_page.html
```diff
@@ -10,7 +10,7 @@
             <div class="grid">
                 <div class="c-1-2-l">
                     <div class="learn__single-category-header-content">
-                        <h1 class="learn__single-category-title h-xl text-blue-deep-80" id="topic-title">{{ page.title }}</h1>
+                        <h1 class="learn__single-category-title h-xl text-blue-deep-80" id="topic-title">{{ page.title }} hello</h1>
```
webbyfox, line 13: is this `hello` required?

exportplan/helpers.py
```diff
@@ -211,3 +212,15 @@ def get_check_duties_link(exportplan):
     # TODO Once requirements have been defined pick country code from export plan
     url = 'https://www.check-duties-customs-exporting-goods.service.gov.uk/'
     return url
+
+
+def get_all_lesson_details():
+    lessons = {}
+    for lesson in models.DetailPage.objects.all():
```
webbyfox, line 219: worth to check only live pages? `...objects.live()`

bobmeredith: Rather than build up this list, could you use `lesson = models.DetailPage.filter(slug=slug)` or `get(slug=slug)` with a try-catch to get the lesson you need? That could go directly in the templateTag perhaps.
My concern would be that 'slug' is not unique, but the map method requires that slug is unique anyway.

bobmeredith (approved):
LGTM

PR #344, opened 2020-09-03
Added Hero block to Lesson

webbyfox:
 - [x] Changelog entry added.


learn/templates/learn/hero_media_item.html
```diff
@@ -1,5 +1,12 @@
 {% load wagtailimages_tags %}
 
-{% image page.image original as hero_image %}
+{% for block in page.hero %}
```
bobmeredith, line 3: I think it would be neater to define the rendering templates in the block definitions? Then, you wouldn't need to switch on the block type. The additional template tag load would also move into the block template.

bobmeredith (changes requested):
I tried this branch out and the migration went ok.  I tested the page with image and video too.

learn/templates/learn/hero_media_item.html
webbyfox: Moved template definition to block

bobmeredith: Super.  I think that's much nicer.

bobmeredith (approved):
LGTM

PR #353, opened 2020-09-08
GP2-420-get-topic-details

depsiatwal:
 - [x] Changelog entry added.


core/migrations/0032_auto_20200908_1151.py
```diff
@@ -0,0 +1,28 @@
...
...icon='fa-play', template='core/personalised_page_struct_video_block.html')), ('content_module', core.blocks.ModularContentStaticBlock()), ('Step', wagtail.core.blocks.StructBlock([('title', wagtail.core.blocks.CharBlock(max_length=255)), ('body', wagtail.core.blocks.RichTextBlock()), ('image', wagtail.images.blocks.ImageChooserBlock(required=False))], icon='cog')), ('fictional_example', wagtail.core.blocks.StructBlock([('fiction_body', wagtail.core.blocks.RichTextBlock(icon='openquote'))], icon='fa-commenting-o', template='learn/fictional_company_example.html')), ('ITA_Quote', wagtail.core.blocks.StructBlock([('quote', wagtail.core.blocks.RichTextBlock()), ('author', wagtail.core.blocks.CharBlock(max_length=255))], icon='fa-quote-left')), ('pros_cons', wagtail.core.blocks.StructBlock([('pros', wagtail.core.blocks.StreamBlock([('item', wagtail.core.blocks.StructBlock([('item', wagtail.core.blocks.CharBlock(max_length=255))], icon='fa-arrow-right'))])), ('cons', wagtail.core.blocks.StreamBlock([('item', wagtail.core.blocks.StructBlock([('item', wagtail.core.blocks.CharBlock(max_length=255))], icon='fa-arrow-right'))]))], icon='fa-arrow-right', template='learn/pros_and_cons.html'))]),
```
bobmeredith, line 26: Just wondering why this streamfield is being altered. AFAICT, you didn't change the model.  Maybe it's migration crossover.

core/wagtail_hooks.py
```diff
@@ -59,3 +59,15 @@ def set_read_time(request, page):
...
+@hooks.register('after_edit_page')
+def set_lesson_pages_topic_id(request, page):
+    if hasattr(page, 'topics'):
+        for topic in page.topics:
+            for empty_topic_page in [
+                topic_pages.specific for topic_pages in topic.value['pages'] if not topic_pages.specific.topic_block_id
+            ]:
+                empty_topic_page.topic_block_id = topic.id
+                empty_topic_page.save()
+        return page
```
bobmeredith, line 73: I think I can see some issues with this.
As you're only looping pages in the topics in this `curated_list_page`, if a page is removed from a topic - or indeed if the lesson page in a topic is changed, the `topic_block_id` won't be cleared.
Also, why the check for `not topic_pages.specific.topic_block_id`? I think you should set the `topic_block_id` even if it's already populated but does not = this `topic.id` (in case a lesson is moved from one topic to another).

You could do this by clearing all lesson pages with this topic_block id `Detail_page.objects.filter(topic_block_id=topic.id)` or - better by making a map of the ids, ticking them off as you find lesson pages in this topic, then wiping the topic_block_id from the remainder.

You're right - it's a tricky thing!

bobmeredith (changes requested):

core/migrations/0032_auto_20200908_1151.py
stevejalim: Yeah, I agree. As a separate ticket, we could add a small test to the suite to catch a missing migration to avoid it in the future.

bobmeredith (approved):
Nice one.

PR #355, opened 2020-09-09
Override alt property for image to use custom alt_text

webbyfox:
 - [x] Changelog entry added.


tests/unit/core/test_models.py
```diff
@@ -1,17 +1,26 @@
...
+    ListPage,
+)
+from domestic.models import DomesticDashboard, DomesticHomePage
 from exportplan.models import ExportPlanDashboardPage
-from domestic.models import DomesticHomePage, DomesticDashboard
 from tests.unit.core import factories
 from .factories import DetailPageFactory
-from wagtail_factories import ImageFactory
-
-from django.core.exceptions import ValidationError
```
stevejalim, line 14: Thanks for cleaning up the import ordering 💅

tests/unit/core/test_models.py
```diff
@@ -111,3 +120,24 @@ def test_validation_kick_for_multiple_hero_image(self):
...
+
+
+class TestImageAltRendition(TestCase, WagtailTestUtils):
+    def setUp(self):
+        self.login()
+        root_collection = Collection.objects.create(name='Root', depth=0)
+        great_image_collection = root_collection.add_child(name="Great Images")
+
+        # Create an image with alt text
+        AltTextImage = get_image_model()
```
stevejalim, line 132: (Not a change request 😄 )

It'd be good to improve this naming of `AltTextImage` before things go too far - `GreatImage` or something more generic will be less misleading if the `AltTextImage` also ends up having, for example, a custom `caption` field too

stevejalim (approved):
LGTM! Thanks for the solution.

One nit and one comment - but neither is a blocker

tests/unit/core/test_models.py
webbyfox: Yeah, Agreed. although any model name changes I would like to do as separate PR/ticket

bobmeredith (approved):
LGTM

PR #359, opened 2020-09-14
GP2-581_GP2-582_GP2_586_spinner_scroll_item_interaction

bobmeredith:

 - [x] Changelog entry added.


core/sass/components/_product-finder.scss
```diff
@@ -6,4 +14,32 @@
...
+  		@include stretch;
+  	}
+  }
+  .shim {
+  	z-index: 3;
+  	@include stretch;
+  	text-align: center;
+  	padding-top: 30%;
+  	&:before {
+  		content: '';
```
dahfool, line 31: not sure if it's github not displaying properly, but it looks like 4 spaces here.

bobmeredith: Good spot - there are tabs in there.
I'll prettify.

dahfool (approved):

PR #370, opened 2020-09-15
(GP2-449) Extended Video model for transcript text

webbyfox:

 - [x] Changelog entry added.


core/models.py
```diff
@@ -27,11 +29,28 @@
...
+    def sources(self):
+        return [{
+            'src': self.url,
+            'type': mimetypes.guess_type(self.filename)[0] or 'application/octet-stream',
+            'transcript': self.transcript
+        }]
+
+
+class GreatMedia(GreatVideo):
+    admin_form_fields = getattr(Media, 'admin_form_fields') + ('transcript', )
```
stevejalim, line 51: Nit: Any reason not just to use `Media.admin_form_fields` here, consistent with `AltTextImage`

stevejalim (commented):
Thanks for this! Makes sense to me, but just wondering if we can simply have a single GreatMedia class and drop the GreatVideo class. Happy to discuss if I've got the wrong idea here

Also got a couple of nits, if you feel like changing those - no worries.

❤️    for the tests!

core/models.py
```diff
@@ -27,11 +29,27 @@
...
 from wagtail_personalisation.models import PersonalisablePageMixin
+from wagtailmedia.models import Media
 
 from core import blocks as core_blocks, mixins
 from core.context import get_context_provider
 
 
+class GreatMedia(Media):
+
+    transcript = models.TextField(verbose_name=_('transcript'), blank=True, null=True)
```
stevejalim, line 40: Tiny thing, missed it first time around.

```suggestion
    transcript = models.TextField(verbose_name=_('Transcript'), blank=True, null=True)
```

If you change it here, edit the migration too so that you don't have to regenerate another one

core/migrations/0033_greatmedia.py
```diff
@@ -0,0 +1,27 @@
...
+        ('wagtailmedia', '0003_copy_media_permissions_to_collections'),
+        ('core', '0032_auto_20200908_1151'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='GreatMedia',
+            fields=[
+                ('media_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='wagtailmedia.Media')),
+                ('transcript', models.TextField(blank=True, null=True, verbose_name='transcript')),
```
stevejalim, line 19: Tiny change with a capital letter - see comment on the model below

```suggestion
                ('transcript', models.TextField(blank=True, null=True, verbose_name='Transcript')),
```

tests/unit/core/test_models.py
```diff
@@ -2,12 +2,16 @@
 
 import pytest
 from django.core.exceptions import ValidationError
+from django.core.files import File
+from django.core.files.base import ContentFile
 from django.test import TestCase
+from six import b
```
stevejalim:
```suggestion
```
We can drop this import now I think

stevejalim (approved):
Thanks for this - looks good to me!

There's one TINY change I didn't spot the first time, just to do with the verbose name for the `transcript` field. Otherwise, definitely LGTM.

Also, the `GreatMedia` naming choice is a good one - it's clear it's custom, but not too specific in the way that AltTextImage is very specific 👍

core/models.py
```diff
@@ -27,11 +29,28 @@
...
+    @property
+    def sources(self):
+        return [{
+            'src': self.url,
+            'type': mimetypes.guess_type(self.filename)[0] or 'application/octet-stream',
+            'transcript': self.transcript
+        }]
+
+
+class GreatMedia(GreatVideo):
```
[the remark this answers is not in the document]
webbyfox: Good spot. I dont know why I missed that inheritance

PR #372, opened 2020-09-18
GP2-512

dahfool:

 - [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-512 


bobmeredith (approved):
LGTM

PR #374, opened 2020-09-18
GP2-500: Include video duration in read time for the page

stevejalim:
This changeset adds counting of video duration in the read time of a page.

Given we're already rendering and parsing the HTML page, the most effective solution was simply to annotate the `<video>` blocks with a custom, non-clashing data attribute, which we then pull out of the HTML with BeautifulSoup and add to our `readtime` calculation.

 - [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-500 

* Have at least one lesson in place - note what the current readtime of it is.
* Add a video to the Media section of the CMS, including a duration in seconds - eg 120
* Edit a lesson and add a video as the Hero block (videos and text in the Personalised content section are ignored for readtime purposes)
* Save that page AS DRAFT.
* Preview the draft and confirm the readtime has generally increased by the equivalent number of minutes
* View the live page and confirm the readtime is still the original value there (ie, we've not updated the live page yet)
* Publish your draft page
* Re-view the live page and confirm the readtime is still now the updated value

Bonus: check direct-to-live change
* In the Media section of the CMS, pick the same video and double its duration
* Edit the lesson page again and simply re-publish it (ie, create a new live version with no other content changes apart from what you've changed for the Video duration over in the media panel)
* View the live page and confirm the readtime has increased by the appropriate amount.

Caveats: 
* If you update the duration on a media file, you need to re-publish the page that features it to be able to update its readtime. Product (Miranda) knows this and seems fine with the limitation
* We still say 'X mins read' on the page, rather than 'X mins' - at least for now, until UX has a clear decision on it

Notes:
* This is my first PR here, and I'm very open to all the feedback I can get on it, to help me get in tune with the team. Don't hold back!
* It's taken longer than hoped, but I may still try to slide in some more tests if you agree they are worthwhile.

### DIFF
--- a/core/wagtail_hooks.py
+++ b/core/wagtail_hooks.py
@@ -1,17 +1,21 @@
-from great_components.helpers import add_next
-from wagtail.core import hooks
+import datetime
+import json
 
-from django.urls import reverse
-from django.shortcuts import redirect
-
-from core import mixins, views
-
-from django.template.loader import render_to_string
-from bs4 import BeautifulSoup
 from datetime import timedelta
+
+from bs4 import BeautifulSoup
 import readtime
 
-from core import models
+from great_components.helpers import add_next
+from wagtail.core import hooks
+from wagtail.core.models import Page
+
+from django.urls import reverse
+from django.core.serializers.json import DjangoJSONEncoder
+from django.shortcuts import redirect
+from django.template.loader import render_to_string
+
+from core import constants, mixins, views, models
 
 SESSION_KEY_LESSON_PAGE_SHOW_GENERIC_CONTENT = 'LESSON_PAGE_SHOW_GENERIC_CONTENT'
 exportplan_templates = ['exportplan/automated_list_page.html', 'exportplan/dashboard_page.html']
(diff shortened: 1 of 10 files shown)


core/wagtail_hooks.py
```diff
@@ -50,17 +54,62 @@ def login_required_signup_wizard(page, request, serve_args, serve_kwargs):
...
+
+    if not page.has_unpublished_changes:
+        # 2. The live version is based on the latest revision, which means
+        # we also need to update the live page, because it won't automatically
+        # reflect the chages we made to that revision.
+
+        # This update()-based approach is awkward but we want to update the
+        # Page record without any side effects via save() etc
+        queryset_for_page = type(page).objects.filter(id=page.id)
+        queryset_for_page.update(**data_to_update)
```
stevejalim, line 88: This function is implicitly tested by the main tests, and because I had to timebox this, I've stuck with that approach for now. However, do say if you feel like this needs specific unit-test coverage and I'll make a ticket to add it

core/wagtail_hooks.py
```diff
@@ -50,17 +54,62 @@ def login_required_signup_wizard(page, request, serve_args, serve_kwargs):
...
+
+
 @hooks.register('after_edit_page')
```
stevejalim, line 91: I appreciated all the discussion around `after_edit_page` vs `after_publish_page` but in the end went with a solution that keeps readtime up to date on both drafts and live pages, which should be better for content editors

core/wagtail_hooks.py
```diff
@@ -50,17 +54,62 @@ def login_required_signup_wizard(page, request, serve_args, serve_kwargs):
...
+        )
+        watching_seconds = sum([
+            int(node.get(constants.VIDEO_DURATION_DATA_ATTR_NAME, 0)) for node in video_nodes
+        ])
+        seconds = reading_seconds + watching_seconds
+
+        _update_data_for_appropriate_version(
+            page=page,
+            data_to_update={'estimated_read_duration': timedelta(seconds=seconds)}
+        )
```
stevejalim, line 112: Because `save_revision()` was a generating and updating a new draft of the page, we had to move to a way that kept the live page (if it existed) as well as any draft up to date with the appropriate readtime.

tests/helpers.py
```diff
@@ -23,3 +28,27 @@ def reload_urlconf(urlconf=None):
         reload(sys.modules[urlconf])
     else:
         import_module(urlconf)
+
+
+def make_test_video(
```
stevejalim, line 33: This was refactored out from work Riz had already done - i just made it reusable 👍

depsiatwal (approved):
LGTM

PR #375, opened 2020-09-21
GP2-500: Add additional test to confirm `_update_data_for_appropriate_version` operates as expected

stevejalim:
This will be a handy regression test.

Note that the off-by-one result at the end of the test is unrelated to the function being tested.

 - [ ] [Changelog](CHANGELOG.md) entry added. - NOT ADDED BECAUSE EXTENSION TO EXISTING FEATURE TESTS


depsiatwal (approved):
LG

PR #379, opened 2020-09-22
Gp2 506 product change confirmation message

webbyfox:
- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/506


bobmeredith (commented):
Nice thing but it would be nicer still to make the confirmation more reusable.

bobmeredith (approved):
Nice one

PR #382, opened 2020-09-23
GP2-542-CC-Country-store-pairs

depsiatwal:
- [x] Ticket exists in Jira https://URL/browse/GP2-542-CC-Country-store-pairs


exportplan/serializers.py
```diff
@@ -50,9 +50,19 @@ class AdaptationTargetMarketSerializer(serializers.Serializer):
     uk_customs_declaration = serializers.CharField(required=False, allow_null=True)
 
 
+class ExportPlanCountrySerializer(serializers.Serializer):
+    country_name = serializers.CharField(required=True)
+    country_iso2_code = serializers.CharField(required=False, allow_null=True)
+
```
bobmeredith, line 56: Question - why allow a country without an iso code? Would it be to cover some existing functionality?

bobmeredith (approved):

PR #383, opened 2020-09-23
GP2-394 - Enter verification code

rafa-garcia:
To do:

 - [x] Ticket exists in Jira https://URL/browse/GP2-394


core/templates/core/signup.html
```diff
@@ -11,8 +11,12 @@
     {{ block.super }}
     <script type="text/javascript">
 		var element = document.getElementById('sign-up');
+		var currentStep = '{% if request.GET.verify %}verification-code{% else %}credentials{% endif %}';
+		var email = '{% if request.GET.verify %}{{request.GET.verify}}{% endif %}';
```
bobmeredith, line 15: I don't think you need the {% if %} here.  Django defaults to empty string if the path doesn't exist.

bobmeredith (approved):
I left a tiny comment
Looks good though.

PR #389, opened 2020-09-24
update dashboard headers

dahfool:


lewis-coulson-dit (approved):

stevejalim (approved):
Looks legit - can't confirm whether it's the expected UI result though

bobmeredith (commented):
Change log?

PR #396, opened 2020-09-25
Update position of the sidebar and textarea padding fixes

dahfool:


core/sass/exportplan/_builder.scss
```diff
@@ -12,7 +12,7 @@
     background-color: $white;
     box-shadow: 0 2px 10px 0 rgba(0,0,0,0.1);
     position: absolute;
-    top: 120px;
+    top: 137px;
```
bobmeredith, line 15: This seems a tad literal.
Is it to adjust for the height of the header?
If so, I think the sidebar is inside the section below the header. If so, give the section a 'position:relative' and then the top would be 0.

bobmeredith: Needs more work, but merge for now

bobmeredith (approved):
Cool

PR #401, opened 2020-09-28
GP2-627 - Fix button styling issues across all pages

lewis-coulson-dit:


react-components/src/components/ProductFinder/CountryFinder.jsx
```diff
@@ -31,7 +31,6 @@ export function CountryFinder(props) {
   const [searchStr, setSearchStr] = React.useState()
```
dahfool, line 31: `useState` is imported at line 1, you can use useState() rather than React.useState()

react-components/src/components/ProductFinder/CountryFinder.jsx
```diff
@@ -65,8 +64,8 @@ export function CountryFinder(props) {
       // map regions
       let regions = {}
       for (const [index, country] of result.entries()) {
-        let region = country.region;
-        (regions[region] = regions[region] || []).push(country)
+        let region = country.region
```
dahfool, line 67: Semi-colon might need to be moved back to line 67.

bobmeredith (approved):
Looks good

react-components/src/components/ProductFinder/CountryFinder.jsx
bobmeredith: I think it's OK. I'm guessing was removed by a prettifier.

react-components/src/components/ProductFinder/CountryFinder.jsx
bobmeredith: True, but not really much of an issue I would have thought.

bobmeredith: Actually, good spot but can we merge this as-is and I'll update on next mod to this component.

PR #405, opened 2020-09-28
add second learn linking component to about your business page

dahfool:


bobmeredith (approved):
Looks OK. I guess a changelog entry not needed

PR #416, opened 2020-09-30
GP2-602 Logout and header across all pages

bobmeredith:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-602


bobmeredith:
Logged in 
image

Menu
image

Logged out
image

Logged out menu
image

bobmeredith:
Testing - the new style header should appear on all pages apart from the homepage and login/signup pages.
It should be present when signed-in or out - but the personalization bar should only appear when signed in.
The top-right menu differs between logged-in and logged out state. 
If the user logs out when on the dashboard page, they will be returned to the login page (as the dashboard is only available when logged in), otherwise, they will stay on the same page.

lewis-coulson-dit (approved):
FE LGTM!

react-components/src/components/ModalMenu.jsx
```diff
@@ -36,23 +36,19 @@ export function Menu(props) {
     modalContent.style.opacity = '1'
   }
 
-  let avatar = props.avatar ? <img src={props.avatar} /> : ''
+  const logout = () => {
+    Services.logout().then((response) => {
```
dahfool, line 40: Might need a catch here, in case the logout fails

bobmeredith: Maybe - I couldn't work out what to do if logout fails though. I mean the user wants to logout - and it's not much good telling him he can't, so we will be refreshing the screen whatever.

bobmeredith: Actually, fetch won't reject unless there is a network failure - I may as well add a catch and refresh the page.

core/wagtail_hooks.py
```diff
@@ -93,9 +93,10 @@ def set_read_time(request, page):
     if hasattr(page, 'estimated_read_duration'):
         html = render_to_string(page.template, {'page': page, 'request': request})
         soup = BeautifulSoup(html, 'html.parser')
-        for tag in soup.body.find_all(['script', 'noscript', 'link', 'style', 'meta']):
+        for tag in soup.body.find_all(['script', 'noscript', 'link', 'style', 'meta', 'header']):
```
stevejalim, line 96: 👏  Nice solution - I definitely agree that we don't need to count page furniture in the readtime.

sso/helpers.py
```diff
@@ -29,8 +29,10 @@ class CreateUserException(APIException):
 
 
 def set_cookies_from_cookie_jar(cookie_jar, response, whitelist):
+    cookie_domain = ''
     for cookie in cookie_jar:
         if cookie.name in whitelist:
+            cookie_domain = cookie_domain or cookie.domain
```
stevejalim, line 35: It looks like we're only ever going to assign a value to `cookie_domain` (or leave it as `''` if the cookie jar is empty, but then we don't do anything with that `cookie_domain` var. What part of the picture am I missing, I wonder?

sso/views.py
```diff
@@ -22,14 +24,35 @@ def post(self, request):
...
 
+class SSOBusinessUserLogoutView(generics.GenericAPIView):
+
+    def post(self, request):
+
+        sso_session_cookie_domain = request.session.get(SSO_COOKIE_DOMAIN_NAME_KEY, '')
+
+        # Call logout on directory_sso to kill the token.
+        upstream_response = requests.post(url=settings.SSO_PROXY_LOGOUT_URL, allow_redirects=False)
+        # Nothing we can do if that fails
```
stevejalim, line 47: Question: is it worth zapping the SSO cookie anyway, regardless of whether they have succesfully killed their SSO session? Or is this the point you made in the meeting that the existence of the SSO-service session will just 'resurrect' their great-cms session anyway? If it is, then I think this is fine -- the user can't log out for reasons beyond great-cms's control, so let's keep them logged in until they DO manage to log out 👍

tests/unit/sso/test_views.py
```diff
@@ -51,6 +55,14 @@ def test_business_sso_user_create_validation_error(client):
     assert response.status_code == 400
 
 
+@pytest.mark.django_db
+def test_business_sso_logout(client, requests_mock):
+    requests_mock.post(settings.SSO_PROXY_LOGOUT_URL, status_code=302)
+    response = client.post(reverse('sso:business-sso-logout-api'), {})
+
+    assert response.status_code == 200
```
stevejalim, line 63: General question as I'm missing a step here: if the mock is returning 302 for the `SSO_PROXY_LOGOUT_URL`, where's the 200 coming from?

stevejalim (commented):
LGTM - one specific question and a couple of general questions, but should be nothing blocking. Happy to approve after the specific question is covered. 

Nice one!

sso/views.py
bobmeredith: Well that's a big question - to which I don't have a hard-and-fast answer.  I'm assuming errors on logging out will be as rare as hen's teeth so I was thinking it's better to get the user to re-try the logout which at least gives us another crack at killing the sso token. If the logout fails several times - I guess the user is just going to close the tab - I mean how many users will actually bother logging off at all?

tests/unit/sso/test_views.py
bobmeredith: Ahha - the mock is for the directory-sso API call - for which 'success' is a 302 redirect (supposedly to a logged-in page) The test is making a call to the logout handler in great-cms which calls the directory-sso mock, gets a 302 and returns a successful 200.
There is a similar test for login with the same odd-looking logic.

sso/helpers.py
bobmeredith: Good spot - that line is redundant. I'll clean it up and PR

sso/views.py
stevejalim: Yeah, I'm definitely with you on this 👍

PR #424, opened 2020-10-01
GP2-451: Case study component

stevejalim:
This PR contains backend and frontent for the Case Study feature mentioned in the ticket.

The backend work:

* Adds a new CaseStudy snippet, following the spec in the ticket, which can be added to a DetailPage just like the ContentModule used to be able to
* Adds two custom tag types, specifically for personalisation - HS-code tags and Country code tags (both just strings)
* Uses these new custom tag types on the CaseStudy snippet, so it's possible to select a case study via these tags
* Removes the ContentModule from the DetailPage's Streamfield, so it won't appear any more
* Adds temporary/placeholder/good-enough-to-test-drive logic to support filtering for a Case Study by HS code (via `?hs-tag=XXXXX`) or country code (via `?country-tag=YYYYY`) or both (via `?hs-tag=XXX&country-tag=YYYY`)
* Roughs in FE content for @lewis-coulson-dit to work with

The frontend work:

* Builds out the new case study module using components from storybook
* (more detail to come from @lewis-coulson-dit if needed)

## To test drive

* Add a new Snippet, picking the CaseStudy option
  * Confirm that to save it, you need to have exactly the following (please play around): 
    * company name
    * summary paragraph
    * one media block in the body, with at least one image or video, and max two. 
    * content in the rich-text area
* Tags are not needed to save the Snippet, but confirm you can add them to both tag fields. Add a couple, save the page and re-edit. Confirm that tags created for one field do NOT autocomplete for the other field.
* Save two snippets - one with a video and an image. Give BOTH Snippets a HS code of `HS1234`, give one of these snippets a country code of DE and save it and another Snippet with a country code of FR and save it. Remember which you saved last.
* Add the CaseStudy component to a DetailPage/lesson page and save it.

**To confirm the temporary tag logic is working:** 
* Browse to that detail page. Add ?hs-tag=HS1234 to the URL - you should see the case study which was saved most recently, including its media content
* Change the querystring to ?country-tag=FR and you should see the case study associated with that tag; now try DE instead of FR
* Note that providing both querystrings at once will not work right now, but that is OK because this logic will be replaced by a proper algorithm very soon.

## Checklist

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-451 
 - [ ] [Changelog](CHANGELOG.md) entry added. --> TO COME AFTER FRONTEND

 - [ ] (if is a UI change) Includes screenshot(s) - ideally before and after, but at least after
 
## Screenshots (Backend)

Screenshot 2020-10-01 at 15 59 54
Screenshot 2020-10-01 at 15 59 57
Screenshot 2020-10-01 at 16 03 47
Screenshot 2020-10-01 at 16 09 09
Screenshot 2020-10-01 at 16 09 28
Screenshot 2020-10-01 at 16 45 18

Screenshot 2020-10-07 at 09 32 27
Screenshot 2020-10-07 at 09 33 02
Screenshot 2020-10-07 at 09 46 49
Screenshot 2020-10-07 at 09 46 54


core/blocks.py
```diff
@@ -224,3 +226,52 @@ class Meta:
...
+    class Meta:
+        admin_text = (
+            'Case Studies are automatically displayed based on '
+            'personalisation rules; no configuration needed beyond '
+            'adding this block to the page.'
+        )
+        icon = 'fa-book'
+        template = 'core/case_study_block.html'
+
+    def _annotate_with_case_study(self, context):
```
stevejalim, line 242: This is where I think the real logic for determining who gets to see this case study will slot in.

I can add tests for it to keep codecov happy, but I suspect it's going to be thrown away soon in favour of the real algorithm

stevejalim: @webbyfox Here, instead of getting the tags from request.GET, we can look at context['request'].user and then run the personalisation logic to get the right tags. Once you have the tags, the code that does the CaseStudy.objects.filter will need a change (at the moment it’s an OR not an AND query, just to show the new HS and country tags working)

core/models.py
```diff
@@ -608,3 +616,170 @@ class ContentModule(ClusterableModel):
...
+        verbose_name='HS-code tags'
+    )
+
+    country_code_tags = ClusterTaggableManager(
+        through='core.CountryTaggedCaseStudy',
+        blank=True,
+        verbose_name='Country tags'
+    )
+
+    created = CreationDateTimeField('created', null=True)
```
webbyfox, line 751: Any reason behind `null=True`. Always,  there should be value for the `created` field.

webbyfox (commented):
Overall, looks good. Worth to review WT data migration for this. Cases where some tags/video do not exist on the target server and it gets migrated via WT process.

core/models.py
stevejalim: True, this is a brand-new model. I took this from other models, but realise it was probably added after the original migrations.

stevejalim:
@webbyfox 
> Overall, looks good. Worth to review WT data migration for this. Cases where some tags/video do not exist on the target server and it gets migrated via WT process.

YES, exactly - you read my mind. I have already added it to my checklist for my WagtailTransfer work, but am super happy we're both thinking about it. 🚀

core/models.py
```diff
@@ -608,3 +616,222 @@ class ContentModule(ClusterableModel):
...
+                    # implicitly, [0] must be an image
+                    # video after image: not allowed
+                    error_messages.append(
+                        'The video must come before a still image.'
+                    )
+
+    return error_messages
+
+
+def case_study_body_validation(value):
```
stevejalim, line 716: @webbyfox Heya Riz - could you please re-review this function and the two private functions it uses - they are the only additions since you reviewed the draft PR for backend code. Thanks!

core/models.py
```diff
@@ -608,3 +616,222 @@ class ContentModule(ClusterableModel):
...
+                if set(subnode_block_types) == {VIDEO_BLOCK}:
+                    # Two videos: not allowed
+                    error_messages.append(
+                        'Only one video may be used in a case study.'
+                    )
+                elif subnode_block_types[1] == VIDEO_BLOCK:
+                    # implicitly, [0] must be an image
+                    # video after image: not allowed
+                    error_messages.append(
+                        'The video must come before a still image.'
```
webbyfox, line 710: @stevejalim  confused with this comment, as the comment above [L708] says `video after image: not allowed` which contradict this line

webbyfox (approved):
Looking really good. @bobmeredith  @dahfool might want to have a look on FE code

dahfool (approved):

PR #429, opened 2020-10-02
ablity-to-disable-export-sections 

depsiatwal:


webbyfox (approved):
LGTM

core/models.py
```diff
@@ -519,7 +519,7 @@ def _export_plan_url_map(self):
         Export Plan sections we have."""
 
         return {
-            entry['url']: entry['title'] for entry in EXPORT_PLAN_SECTION_TITLES_URLS
+            entry['url']: entry['title'] for entry in list(EXPORT_PLAN_SECTION_TITLES_URLS.values())
```
stevejalim, line 522: Nit: do we def need the `list()` here? I think `odict_values`  from `values()` is iterable.

exportplan/data.py
```diff
@@ -1,30 +1,93 @@
...
+ABOUT_YOUR_BUSINESS = 'About your business'
+OBJECTIVES = 'Objectives'
+TARGET_MARKETS_RESEARCH = 'Target markets research'
+ADAPTATION_TARGET_MARKET = 'Adaptation for your target market'
+MARKETING_APPROACH = 'Marketing approach'
+COSTS_AND_PRICING = 'Costs and pricing'
+FINANCE = 'Finance'
+PAYMENT_METHODS = 'Payment methods'
+TRAVEL_AND_BUSINESS_POLICIES = 'Travel and business policies'
+BUSINESS_RISK = 'Business risk'
```
stevejalim, line 16: ❤️

exportplan/data.py
```diff
@@ -1,30 +1,93 @@
...
+    SECTIONS_DISABLED = [
+        TARGET_MARKETS_RESEARCH,
+        ADAPTATION_TARGET_MARKET,
+        MARKETING_APPROACH,
+        COSTS_AND_PRICING,
+        FINANCE,
+        PAYMENT_METHODS,
+        TRAVEL_AND_BUSINESS_POLICIES,
+        BUSINESS_RISK,
+    ]
```
stevejalim, line 30: I get your point from our call the other day that this'll need a code change and release to enable/disable a part of the plan, but still agree that without putting this all in the DB or a separate config file, this is a reasonable balance/compromise. 

If we find people want to flip EP content a lot, maybe we could use a DB-based feature flipper like https://waffle.readthedocs.io/en/stable/

exportplan/data.py
```diff
@@ -1,30 +1,93 @@
...
+else:
+    SECTIONS_DISABLED = []
 
-SECTION_URLS = [reverse_lazy('exportplan:section', kwargs={'slug': slug}) for slug in SECTION_SLUGS]
+SECTIONS = OrderedDict({
+    slugify(ABOUT_YOUR_BUSINESS): {
+        'title': ABOUT_YOUR_BUSINESS,
+        'url': reverse_lazy('exportplan:section', kwargs={'slug': slugify(ABOUT_YOUR_BUSINESS)}),
+        'disabled': True if ABOUT_YOUR_BUSINESS in SECTIONS_DISABLED else False,
```
stevejalim, line 38: Nit: could maybe move this into a reusable func that just takes the section's constant and returns a Boolean, but it's no big deal. Easy for me to say in review, as I didn't have to battle through the refactor :)

stevejalim: Indeed, given how often we have to repeat the relevant constant for each entry, we'd be at a slightly lower risk of a copy-paste error, or the like, if the key and dict value for each item was generated by a helper func.

However, I don't think we _need_ to change anything here

exportplan/data.py
```diff
@@ -1,30 +1,93 @@
...
+        FINANCE,
+        PAYMENT_METHODS,
+        TRAVEL_AND_BUSINESS_POLICIES,
+        BUSINESS_RISK,
+    ]
+else:
+    SECTIONS_DISABLED = []
 
-SECTION_URLS = [reverse_lazy('exportplan:section', kwargs={'slug': slug}) for slug in SECTION_SLUGS]
+SECTIONS = OrderedDict({
```
stevejalim, line 34: +1 for explicitly using an `OrderedDict`, even though regular py3 dict also now keeps order. `OrderedDict` immediately communicates what you're doing here

exportplan/data.py
```diff
@@ -1,30 +1,93 @@
...
-SECTION_TITLES_URLS = [
-    {
-        'title': section,
-        'url': reverse_lazy('exportplan:section', kwargs={'slug': slugify(section)})
-    } for section in SECTION_TITLES
-]
+
+SECTION_TITLES = [val['title'] for val in SECTIONS.values()]
+SECTION_SLUGS = list(SECTIONS.keys())
+SECTION_URLS = list(SECTIONS.values())
```
stevejalim, line 90: Again, same relaxed question re the need for `list()` -- though I guess that the cast to `List` does remove the risk of surprise at the consumer end of things 😄

stevejalim (approved):
Nice one - LGTM! I agree with consolidating data into a single canonical data structure, then pulling them apart as each consumer needs.

I've added some comments from the 'easy-to-say-because-i-didn't-write-it' perspective, but nothing that's a blocker or which I think is massively worth a change. However, if we have to add more sections to the EP, doing things to reduce repetition might be worth it then.

PR #434, opened 2020-10-06
GP2-682_product_search_rework

bobmeredith:

 - [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-682


bobmeredith:
image

image

bobmeredith:
Manual testing.
A range of suitable searches including those with multiple possible items.
e.g. 'cheese' gives a direct follow on question 
'leather belt' gives an item selection to branch between 'belt' and 'leather'

react-components/src/components/ProductFinder/Interaction.jsx
```diff
@@ -0,0 +1,116 @@
...
+import Services from '@src/Services'
+
+
+const capitalize = (str, enable = true) => {
+  const strWithSpaces = str.replace(/_/g, ' ')
+  return enable ? strWithSpaces.charAt(0).toUpperCase() + strWithSpaces.slice(1) : strWithSpaces
+}
+
+function RadioButtons(props) {
+  const { attribute, valueChange } = props;
```
lewis-coulson-dit, line 13: I'd be tempted to set default values here so you don't need to check `attribute.attrs` twice. `const { attribute = {id="", attrs=[]}, valueChange } = props;`

dahfool:
@bobmeredith All good for me execpt disabling -  eslint-disable prefer-destructuring as you are already desconstructure in the functional components, much as well do when you define them and re-enable the linting e.g

export default function ExpandCollapse({
 buttonLabel, 
 defaultExpanded, 
 children
})

react-components/src/components/ProductFinder/Interaction.jsx
```diff
@@ -0,0 +1,116 @@
...
+  const [selection, setSelection] = useState()
+
+  const updateSelection = (_selection) => {
+    setSelection(_selection)
+    valueChange(_selection)
+  }
+
+  useEffect(() => {
+    (attribute.attrs || []).map((option) => {
```
lewis-coulson-dit, line 22: Could save yourself some typing and destructure the option param to - `.map(({value, name, id}) => {` but no biggie

react-components/src/components/ProductFinder/ProductFinder.jsx
```diff
@@ -6,6 +6,19 @@ import ReactModal from 'react-modal'
 import Services from '@src/Services'
 import Spinner from '../Spinner/Spinner'
 import Confirmation from './MessageConfirmation'
+import Interaction from './Interaction'
+import ExpandCollapse from './ExpandCollapse'
+
+
+const capitalize = (str, enable = true) => {
```
lewis-coulson-dit, line 13: Not sure if we have a utils file for this kind of stuff but we probably should.

react-components/src/components/ProductFinder/ProductFinder.jsx
```diff
@@ -203,40 +215,18 @@ function ProductFinder(props) {
...
-                <div className="table-cell bold capitalize">{searchResults.productDescription}</div>
-              </div>
-            </section>
-          )}
+          {searchResults.hsCode && sectionFound(searchResults)}
           {Section('Please choose your item', itemChoice)}
-          {!itemChoice && Section(`Tell us more about your '${searchResults.currentItemName}'`, questions)}
-          {!itemChoice && Section('Your item&apos;s characteristics', known)}
-          {!itemChoice && Section('We\'ve assumed:', assumptions)}
+          {!itemChoice && Section(`Tell us more about "${searchResults.currentItemName}"`, questions)}
```
lewis-coulson-dit, line 226: Can use a React fragment here so you just need to do the check once:

`{!itemChoice && <>
 Section(`Tell us more about "${searchResults.currentItemName}"`, questions)}
...
</>`

react-components/src/components/ProductFinder/ProductFinder.jsx
bobmeredith: Couldn't find one, but there is a 'helpers' file with not much in it.

react-components/src/components/ProductFinder/Interaction.jsx
bobmeredith: That's not really the same thing. The default for 'attribute' would only apply if attribute is undefined which is not going to happen. There is still the possibility of attribute being defined but not having an 'attrs' array.  Actually, I don't know if this is possible, but these data are coming in from an external source so I was playing safe.

react-components/src/components/ProductFinder/Interaction.jsx
bobmeredith: Fair enough. I'm not a huge fan of destructuring from a readability POV, but in this case it's OK.

react-components/src/components/ProductFinder/ProductFinder.jsx
bobmeredith: Yes, this can be reduced now.  Originally, there was specific logic on displaying each section, but the new designs have rather simplified it for now.

react-components/src/components/ProductFinder/Interaction.jsx
lewis-coulson-dit: Ok yea makes sense

react-components/src/components/ProductFinder/ProductFinder.jsx
bobmeredith: OK, moved it to helpers and added a test.

react-components/src/components/ProductFinder/Interaction.jsx
bobmeredith: done

lewis-coulson-dit (approved):

PR #440, opened 2020-10-08
[GP2-688] Implement GTM tags in product search modal

rafa-garcia:
- [x] https://uktrade.atlassian.net/browse/GP2-688
 - [x] Jira ticket has the correct status
 - [x] [Changelog](CHANGELOG.md) entry added
 - [x] This PR can be merged by reviewers


react-components/src/components/ProductFinder/ProductFinder.jsx
```diff
@@ -58,6 +59,16 @@ function ProductFinder(props) {
       })
   }
 
+  const addTagging = () => {
+    if (window.dataLayer) {
```
bobmeredith, line 63: I think there will be a number of events that need to push tags into the datalayer (I assume that's how it works)
As such, a function 'addTagging' seems a bit generic.
I haven't looked at the spec, but I'm guessing that all the events in the product finder will have the same virtualPageUrl/title in which case, I'm thinking you should make the 'event' a parameter.

react-components/src/components/ProductFinder/ProductFinder.jsx
```diff
@@ -30,6 +30,7 @@ function ProductFinder(props) {
   const openModal = () => {
     setProductConfirmationRequired(!!selectedProduct)
     setIsOpen(!selectedProduct)
+    addTagging()
```
bobmeredith, line 33: I think this is in the wrong place.  This function might open the modal, or might just open the confirmation dialogue, in which case, the modal gets opened in line 43. I'm thinking there should maybe be some restructuring to make life easier.

bobmeredith (approved):
LGtM

PR #465, opened 2020-10-13
GP2-756_fix_social_logout_avatar

bobmeredith:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-756
 
We need the domain of the sso_session cookie in order to delete it.  Previous method was to store the domain during login when the cookie is transferred from the directory-sso response to the great-cms response.
This doesn't work for social login however as no great-cms endpoint is called during login and everything is handled through directory-sso.  The solution, use the domain on the display_logged_in cookie instead - that cookie being updated on the response from the logout request, so there's no need to store the domain anymore.


rafa-garcia (approved):
LGTM. Codecov is not happy though

PR #475, opened 2020-10-15
Added test and various tweaks for Country Finder

webbyfox:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/TICKET_ID_HERE 


react-components/src/components/ProductFinder/RegionToggle.jsx
```diff
@@ -1,25 +1,28 @@
-import React, { useState } from 'react'
+import React, { useCallback, useState } from 'react'
```
dahfool, line 1: `useCallback` and `import { forceReRender } from '@storybook/react'` might have been accidently added.

dahfool (approved):

react-components/src/components/ProductFinder/CountryFinder.jsx
```diff
@@ -36,8 +36,10 @@ export function CountryFinderModal(props) {
   }
 
   const searchChange = (evt) => {
-    setSearchStr(evt.target.value.toUpperCase())
-    setExpandRegion(true)
+    if (evt.target.value.length > 0) {
+      setSearchStr(evt.target.value.toUpperCase())
+      setExpandRegion(true)
```
bobmeredith, line 41: I think the expandAll should clear if the search box is empty, rather than force the user to click the 'collapse all' button to close. Also, I think this might break the the removal of the filter if the filter string is removed (since evt.target.value.length == 0)
Suggest remove the conditional and use `setExpandRegion(evt.target.value.length)`

PR #479, opened 2020-10-16
Added auto escape for case study section

webbyfox:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/TICKET_ID_HERE 


core/templates/core/case_study_block.html
```diff
@@ -6,8 +6,8 @@
         <div class="case-study__content media-block">
             <i class="fas fa-newspaper"></i>
             <div>
-                <h3 class="h-m m-b-xs p-0">{{case_study.summary}}</h3>
-                <p class="case-study__company h-xs p-0">{{case_study.company_name}}</p>
+                <h3 class="h-m m-b-xs p-0">{% autoescape off %}{{case_study.summary}}{% endautoescape %}</h3>
```
bobmeredith, line 9: This doesn't feel at all right.  
These fields should be text not HTML, so an apostrophe should get encoded as '&apos;' or something and appear correctly.
Encoding is enabled by default because otherwise you are vulnerable to cross-site-scripting.

react-components/src/components/CaseStudy/CaseStudy.jsx
```diff
@@ -29,10 +29,10 @@ const CaseStudy = ({ content: { heading, company, body } }) => {
       <div className="case-study__content media-block">
         <i className="fas fa-newspaper" />
         <div>
-          <h3 className="h-m m-b-xs p-0">{heading}</h3>
+      <h3 className="h-m m-b-xs p-0" dangerouslySetInnerHTML={{ __html: heading }} />
```
bobmeredith, line 32: Likewise - it's called 'dangerour' for a reason.  There is something wrong here and I don't think this is the solution

bobmeredith (changes requested):

bobmeredith:
I had a look at the code. It looks like the casestudy gets re-rendered by javascript, so the issue is that it's being encoded twice.
The neatest way to fix it is probably similar to the way the body is handled - i.e. in the script block in case_study_block, 
```
company:document.querySelector('#case_study . case-study__company').innerText
```
(If you use innerText, you're getting just text so when React re-renders, it should be html encoded in the normal way.)
If you grab the HTML instead as 'body' is done, it will need the 'dangerouslySetInnerHTML' - but I think it's better to avoid that where possible.

Actually, I think the block of querySelectors would be neater by setting a variable to the case_study ->
```
let caseStudy = document.querySelector('#case_study')
... 
heading: caseStudy.querySelector('h3'.innerText,)
company: caseStudy.querySelector('.case-study__company').innerText,
...
```

webbyfox:
@bobmeredith would you mind to have a look when you got a moment

core/templates/core/case_study_block.html
```diff
@@ -6,8 +6,8 @@
         <div class="case-study__content media-block">
             <i class="fas fa-newspaper"></i>
             <div>
-                <h3 class="h-m m-b-xs p-0">{{case_study.summary}}</h3>
-                <p class="case-study__company h-xs p-0">{{case_study.company_name}}</p>
+                <h3 class="h-m m-b-xs p-0">{{case_study.summary|safe}}</h3>
+                <p class="case-study__company h-xs p-0">{{case_study.company_name|safe}}</p>
```
bobmeredith, line 10: I don't think you should have a safe filter on heading or company. case_study.summary and case_study.company_name are both text, not HTML.

react-components/src/components/CaseStudy/CaseStudy.jsx
bobmeredith, line 32: You're getting innerText, so I think `<h3 className="h-m m-b-xs p-0">{heading}</h3>` is fine. React should escape your apostrophe to `&apos;` at this step.

react-components/src/components/CaseStudy/CaseStudy.jsx
```diff
@@ -29,10 +29,10 @@ const CaseStudy = ({ content: { heading, company, body } }) => {
...
           {isOpen && (
             <>
-              <p className="case-study__company h-s p-0">{company}</p>
+              <p className="case-study__company h-s p-0" dangerouslySetInnerHTML={{ __html: company }} />
               <div className="case-study__media body-l" dangerouslySetInnerHTML={{ __html: body }} />
```
bobmeredith, line 36: Ditto

bobmeredith (changes requested):

bobmeredith (approved):
LGTM

PR #489, opened 2020-10-19
GP2-593_dashboard_tidy

bobmeredith:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-593 


bobmeredith:
image
image
image
image

bobmeredith:
Testing - you need a brand new user.
Initially, all three routing components will be in the top line.
If any lesson set to 'read', the leaning progress module will show and the learning routeing block will disappear.
If the target market is set, the 'add market' route will disappear.
If the export plan dashboard page is visited, the 'export plan' routing section will disappear and a 'view export plan' module will appear in the dashboard.

bobmeredith:
Note, I found that there were no tests on the export plan dashboard page as the factory being used by tests actually produced a lesson detail page instead.
Rather than try to fix all those tests, I renamed the old factory to 'Pseudo' and created a new factory that actually produced an export plan dashboard, along with a view test for it.

core/constants.py
```diff
@@ -1,6 +1,7 @@
 SERVICE_NAME = 'great-cms'
 DASHBOARD_URL = '/dashboard/'
 LOGIN_URL = '/login/'
+EP_DASHBOARD_URL = '/export-plan/dashboard/'
```
depsiatwal, line 4: possible to use full name export_plan_dashboard_ul makes more readable ?

core/templates/core/includes/routing_bar.html
```diff
@@ -8,12 +9,31 @@
 **************
 {% endcomment %}
 
+
+{# Build a list of valid routing blocks #}
```
depsiatwal, line 13: Quite a lot of logic in the template here.
How possible would it be to move to a view where it can be unit tested also.

depsiatwal (commented):
generally lgtm few small comments

core/templates/core/includes/routing_bar.html
bobmeredith: It is.  The logic used to live in the view with tests but it felt wrong living there, given that this is presentation stuff rather than data. At one time, it was tested in the browser tests, but it looks like those tests have been removed.
IMHO, we should be switching away from unit tests to integration/automation tests anyway for this very reason.

depsiatwal: fair enough :-)

depsiatwal (approved):
LGTM

PR #496, opened 2020-10-20
GP2-726: Redirect unauthenticated users on CMS pages to the sign-in page

stevejalim:
This changeset adds`mixins.AuthenticatedUserRequired` to the base page used for most CMS pages (`CMSGenericPage`) so that, if the user is unauthenticated, they are redirected to the login page.

This also has the (requested/desired) side effect that the existing behaviour upon logout (kill session and reload the same page) now causes a redirect when the page is reloaded. Ideally we'd skip the reload, but given this is is a temporary change, I think it's fine for now)

Note that when a user is redirected to signin, we don't remember where they wanted to go originally. This is not ideal, but tolerable within the AC for this piece.

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-726 

### DIFF
--- a/tests/unit/core/test_models.py
+++ b/tests/unit/core/test_models.py
@@ -12,6 +12,7 @@ from wagtail.images.tests.utils import get_test_image_file
 from wagtail.tests.utils import WagtailPageTests, WagtailTestUtils
 from wagtail_factories import ImageFactory
 
+from core.mixins import AuthenticatedUserRequired
 from core.models import (
     AbstractObjectHash,
     CuratedListPage,
@@ -80,7 +81,13 @@ def test_detail_page_anon_user_not_marked_as_read(client, domestic_homepage, dom
 
 
 @pytest.mark.django_db
-def test_curated_list_page_has_link_in_context_back_to_parent(client, domestic_homepage, domestic_site):
+def test_curated_list_page_has_link_in_context_back_to_parent(
+    client,
+    domestic_homepage,
+    domestic_site,
+    user,
+    patch_export_plan
+):
 
     list_page = factories.ListPageFactory(
         parent=domestic_homepage,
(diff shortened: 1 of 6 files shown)


tests/unit/core/test_models.py
```diff
@@ -418,6 +427,44 @@ def test_validation_kick_for_multiple_hero_image(self):
...
+    ]
+
+    for page in pages:
+        assert isinstance(page, AuthenticatedUserRequired)
+
+    for page in pages:
+        response = client.get(page.url, follow=False)
+        assert response.status_code == 302
+        assert response._headers['location'] == ('Location', '/login/')
+
```
bobmeredith, line 460: Would be nice to check the non-authenticated pages too - Homepage and login say

tests/unit/core/test_models.py
```diff
@@ -418,6 +427,44 @@ def test_validation_kick_for_multiple_hero_image(self):
...
+    # Show an authenticated user can still get in there
+    client.force_login(user)
+    for page in pages:
+        response = client.get(page.url, follow=False)
+        assert response.status_code == 200
+
+
```
bobmeredith, line 467: Likewise, could we check that logged-in users get redirected (to dashboard) when they hit the homepage

tests/unit/core/test_models.py
stevejalim: Yep, I get you. I think those are already covered in test_views.py

https://github.com/uktrade/great-cms/blob/develop/tests/unit/core/test_views.py#L135

tests/unit/core/test_models.py
stevejalim: As above

bobmeredith: Fair enough

bobmeredith (approved):
LGTM

PR #506, opened 2020-10-21
Upgraded dependencies packages

webbyfox:


stevejalim (approved):
LGTM as long as `lxml` doesn't have breaking changes

We should use dependabot to make this all easier!

PR #521, opened 2020-10-22
GP2-34: Track module progress on the learning dashboard

stevejalim:
This changeset adds a progress bar to the Learning Dashboard, with module-level completion stats

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-34 

   - Go to the learning dashboard /learn/categories/
   - Shift-click each module to open in a new tab (confirming that the entire module is now clickable, not just the button)
   - Click through to different lessons and mark each as completed
   - Each time, refresh the learning dashboard and confirm that the relevant progress bar increases appropriately
   - Mark every lesson as NOT completed
   - Confirm the learning dashboard now shows 'Get started' instead of a progress button

Screenshot 2020-10-22 at 15 40 33


core/templatetags/object_tags.py
```diff
@@ -6,5 +6,7 @@
 @register.filter
 def get_item(dict, key):
     if hasattr(dict, 'get'):
-        return dict.get(key.lower())
+        if isinstance(key, int):
```
bobmeredith, line 9: You're right - the 'lower' probably isn't needed, and would get rid of the complexity here.

bobmeredith (approved):
I think you could simplify get_item and remove the 'lowercase'. Then it all becomes much simpler.

PR #544, opened 2020-10-26
private@digital.trade.gov.uk -> great.support@trade.gov.uk

logikblok:
Changed email to align to support inbox

image

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-875
 - [X] (if is a UI change) Includes screenshot(s) - ideally before and after, but at least after


stevejalim (changes requested):
Looks good to me, but just spotted a note from Santosh in JIRA

> Can you please add Sign up page as well, email address needs to change.

webbyfox:
shall we add an email address variable in the setting that can be used where ever it needed?

stevejalim:
> shall we add an email address variable in the setting that can be used where ever it needed?

yeah, that's worth doing - a setting and a context processor, perhaps?

webbyfox:
I've added new env to all three envs in vault

bobmeredith (approved):
Looks good to me

stevejalim:
@logikblok Looks good to me, thanks 

But do we also need to act on this from Santosh in the JIRA ticket?

> Can you please add Sign up page as well, email address needs to change.

or does that single template cover Sign up already?

webbyfox:
yes, it's basically a different template but the same code

stevejalim (approved):
LGTM! Thanks for the fixups

PR #558, opened 2020-10-26
Add variables to Universal Data Object

rafa-garcia:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-701
 - [X] This PR can be merged by reviewers


core/templates/core/base.html
```diff
@@ -12,8 +12,17 @@
...
 
+    <script type="text/javascript">
+      var product = '{{ export_plan.export_commodity_codes.0 }}';
+      var country = '{{ export_plan.export_countries.0.country_name }}';
+      var dataLayer = window.dataLayer || []
+      dataLayer.push({
+        'productAdded': product || "False",
+        'marketAdded': country || "False",
+      });
+    </script>
```
bobmeredith, line 24: This doesn't look quite right.  {{ export_plan.export_commodity_codes.0 }} will have a blank value if any of the items in the path doesn't exist. thus `product = ''` - an empty string.
In this case, the JSON pushed will be productAdded: "False"  i.e. the literal string 'False'.
If the product does exist, productAdded would be the object `{commodity_name:<name>, id:<id>}` or somesuch.
I don't think you want either of these things - I expect you want to push a JSON boolean - true or false.
 ```
      var product = '{{ export_plan.export_commodity_codes.0.commodity_name }}';
      var country = '{{ export_plan.export_countries.0.country_name }}';
      var dataLayer = window.dataLayer || []
      dataLayer.push({
        productAdded: !!product,
        marketAdded: !!country,
      });
```

rafa-garcia: Right, I've changed it to get the `commodity_code` (rather than the name) as per requirements 👍 

It's required to return a `commodity_code` and a `country_name` if they're present, otherwise the values should be "False" (literal string).

bobmeredith: OK, fair enough - IMHO, the literal string 'False' is just plain confusing.

bobmeredith (approved):
LGTM

PR #596, opened 2020-10-30
Require anonymous user on specific pages

rafa-garcia:
- [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-954 


core/urls.py
```diff
@@ -12,7 +13,7 @@
 def anonymous_user_required(function):
     inner = user_passes_test(
         lambda user: bool(user.is_anonymous),
-        reverse_lazy('core:login'),
+        constants.DASHBOARD_URL,
```
webbyfox, line 16: This doesn't make sense to me, are we allowing the anonymous user on the dashboard?

rafa-garcia: https://docs.djangoproject.com/en/2.2/topics/auth/default/#django.contrib.auth.decorators.user_passes_test
```
login_url
Lets you specify the URL that users who don’t pass the test will be redirected to. It may be a login page and defaults to settings.LOGIN_URL if you don’t specify one.
```

bobmeredith, line 16: This is well complicated - but valid I think. How about some comments and using named parameters for the call (although 'login_url' is confusing in itself) for the befit of those that follow

bobmeredith (approved):
LGTM

webbyfox (approved):

PR #610, opened 2020-11-03
GP2-884  refactor topic modelling  part two helper code

stevejalim:
Here's another changeset of updates as part of the topic refactor. There are lots more to come but am trying to PR whenever tests are passing to make review easier.

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-884 
 
### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -665,11 +665,7 @@ class DetailPage(CMSGenericPage):
 
     @cached_property
     def topic_title(self):
-        if self.topic_block_id:
-            topic_pages = self.get_parent()
-            for topic in topic_pages.specific.topics:
-                if topic.id == self.topic_block_id:
-                    return topic.value['title']
+        return self.get_parent().title
 
     @cached_property
     def module(self):
--- a/core/models.py
+++ b/core/models.py
@@ -417,19 +417,21 @@ class CuratedListPage(CMSGenericPage):
         StreamFieldPanel('topics')
     ]
 
+    def get_topics(self, live=True) -> models.QuerySet:
+        qs = TopicPage.objects.descendant_of(self)
+        if live:
+            qs = qs.live()
+        return qs
+
     @cached_property
     def count_topics(self):
-        return len(self.topics)
+        return self.get_topics().count()
 
     @cached_property
     def count_detail_pages(self):
         count = 0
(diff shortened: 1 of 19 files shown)


core/models.py
```diff
@@ -417,19 +416,21 @@ class CuratedListPage(CMSGenericPage):
...
     def count_detail_pages(self):
         count = 0
-        for topic in self.topics:
-            for lesson_or_placeholder_block in topic.value.get(
-                'lessons_and_placeholders'
-            ):
-                if lesson_or_placeholder_block.block_type == LESSON_BLOCK:
-                    count += 1
+        for topic in self.get_topics():
+            count += DetailPage.objects.descendant_of(topic).count()
```
webbyfox, line 433: Is there a better way to count detailpages? feels bit odd to count this way.

core/models.py
```diff
@@ -663,17 +664,12 @@ def serve(self, request, *args, **kwargs):
 
     @cached_property
     def topic_title(self):
-        if self.topic_block_id:
-            topic_pages = self.get_parent()
-            for topic in topic_pages.specific.topics:
-                if topic.id == self.topic_block_id:
-                    return topic.value['title']
+        return self.get_parent().title
```
webbyfox, line 667: :heart:

depsiatwal (approved):
On the surface LGTM

core/models.py
webbyfox: just checking if `descendant_of` or similar filter can take a list of pages rather just one?

stevejalim: It can only get the descendant of the specific single page instance, unfortunately.

I know it's a bit odd, but there are no FKs that represent the tree structure to do chained filtering, etc

stevejalim: We could maybe call `DetailPage.objects.descendant_of(module).descendant_of(topic).count()` perhaps

core/models.py
stevejalim: Yeah so many things are much simpler now

webbyfox (approved):

core/models.py
stevejalim: Tried it but no luck. Living with the version here for now

PR #627, opened 2020-11-05
GP2-962: Refactor topics - data migration

stevejalim:
This changeset adds a data migration to re-parent the lesson pages in the system with new topic pages.

For every `CuratedListPage`, add 1...n `TopicPages` as their direct children, then make the appropriate new `TopicPage` as the parent of the existing `DetailPage` (lesson) and/or create a `LessonPlaceholderPage` as needed. 
The source of which lesson/placeholder to put on which topic comes from the `CuratedListPage.topics` `StreamField`

Includes a test of the reparenting function, which includes confirming that the order of the lessons and placeholders is maintained.

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-962 


core/utils.py
```diff
@@ -1,21 +1,27 @@
...
 def get_first_lesson(module):
     """
     Helper function to get first lesson of a module
     @returns: a single DetailPage objects (lesson)
     """
-    return get_all_lessons(module).first()
+    try:
+        return get_all_lessons(module)[0]
+    except IndexError:
+        return None
```
webbyfox, line 24: `.first()` do the same, no?

tests/unit/migrations/test_topics_refactor_data_helpers.py
```diff
@@ -0,0 +1,355 @@
...
+    DetailPage,
+    CuratedListPage,
+    LessonPlaceholderPage,
+    TopicPage,
+)
+
+from core.migration_helpers import create_topics_and_reparent_lessons
+
+
+def _setup_old_heirarchy(domestic_homepage):
```
depsiatwal, line 23: Worth while doing a test and then use following pattern 
To test pre and post states of the data.
i.e Migration shape is as expected
https://github.com/uktrade/directory-forms-api/blob/develop/submission/tests/test_mirations.py

stevejalim: Yeah - thanks for the reminder. Wondered about using it, but I remembered that last time it got blocked from running at all because it couldn't set up state for `core` - possibly related to the squashed migration at the bottom of the migration chain.

core/utils.py
stevejalim: Not now that `get_all_lessons` returns a `list` instead of a `QuerySet`

core/utils.py
```diff
@@ -45,7 +51,7 @@ def get_module_topics(self):
         return self.module.specific.get_topics()
 
     def total_module_topics(self):
-        return self.get_module_topics().count()
+        return len(self.get_module_topics())
```
webbyfox, line 54: `get_module_topics` is queryset so doing `count` is much better than `len` IMHO

stevejalim: Yeah - you're right - was left over from refactoring something else

webbyfox:
Looks good...

webbyfox (approved):

PR #632, opened 2020-11-06
Epic GP2-959: Topics refactor

stevejalim:
This PR contains the result of a number of smaller PRs, all of which were code reviewed before merging to this branch. 

No code has been directly added to this branch


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - GP2-434 - Dashboard routing blocks align
 - GP2-952 - Update tertiary button hover/active styles
 - GP2-681 - HR colour in product finder 
+- GP2-681 - HR colour in product finder
```
webbyfox, line 22: repeated line as one above

core/models.py
```diff
@@ -410,22 +413,24 @@ class CuratedListPage(CMSGenericPage):
     content_panels = CMSGenericPage.content_panels + [
         FieldPanel('heading'),
         ImageChooserPanel('image'),
-        StreamFieldPanel('topics')
+        # StreamFieldPanel('topics')  # TODO: remove me with the field
```
webbyfox, line 416: do not want to remove this?

core/templatetags/content_tags.py
```diff
@@ -60,39 +63,46 @@ def get_backlinked_url(context, outbound_url):
...
-    for topic_block in clp.specific.topics:
-        for lesson_or_placeholder in topic_block.value.get('lessons_and_placeholders', []):
-            if lesson_or_placeholder.block_type == LESSON_BLOCK:
-                if lesson_or_placeholder.value == detail_page:
-                    return topic_block.value.get('title')
-
-    return ''
+    """For the given lesson, find the topic it belongs to and
+    return that topic's title"""
+    return detail_page.get_parent().title
```
webbyfox, line 68: not checking`TopicPage` (detail_page.get_partent) live or not?

stevejalim: I thought about this originally, but to get to this stage, the parent of the DetailPage page _has_ to be live, else you can't reach the DetailPage

core/models.py
stevejalim: It's going to be removed in https://uktrade.atlassian.net/browse/GP2-1020

core/templatetags/content_tags.py
webbyfox: not sure this would be worth to add, though? thinking of the case, if `TopicPage` is not live, and associated  `DetailPage`  is live. Will user will be able to reach detail page?

core/models.py
stevejalim: I'll remove the panel setup when i drop the `topics` field from `CuratedListPage`

core/templatetags/content_tags.py
stevejalim: They won't be able to reach it, because if the topic is not live then it won't be shown on the `curated_list_page` template

webbyfox (approved):
LGTM

PR #633, opened 2020-11-06
Feature/api endpoint check aa 155

agiamas:

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/AA-155 
 

tests/unit/core/test_views.py
```diff
@@ -740,6 +741,33 @@ def test_bad_auth_with_enc_token(client):
...
+    assert res['CCCE_API']['status'] == status.HTTP_200_OK
+    assert res['status'] == status.HTTP_200_OK
+    assert mock_search_commodity_by_term.call_count == 1
+
+
+@pytest.mark.django_db
+@mock.patch.object(helpers, 'search_commodity_by_term')
+def test_check_view_error(mock_search_commodity_by_term, client):
+    # the API is down
+    mock_search_commodity_by_term.return_value = create_response(json_body={'error': 'service unavailable'})
```
bobmeredith, line 762: I'm thinking you should also test http response failures in the service.  You can pass status_code=503/404 to create_response, rather than just check the response body.

core/views.py
```diff
@@ -284,6 +288,19 @@ def get(self, request):
...
+    def get(self, request):
+        response = helpers.search_commodity_by_term(term='feta', json=False)
+        try:
+            response_code = response.json()['data']['hsCode']
+            return Response({'status': status.HTTP_200_OK,
+                             'CCCE_API': {'status': status.HTTP_200_OK, 'response_body': response_code,
+                                          'elapsed_time': math.floor(response.elapsed.total_seconds() * 1000)}})
+        except Exception as e:
+            logger.exception(e)
+            return Response({'status': status.HTTP_200_OK, 'CCCE_API': {'status': status.HTTP_503_SERVICE_UNAVAILABLE}})
```
bobmeredith, line 301: Might be helpful to return (and log) the actual status code returned. rather than forcing a 503. Then we could discriminate between an expired token and the service being down say.
Obviously, we could get to this point on a 200 (if the json was invalid or incorrect) so would need to discriminate between them.

core/views.py
```diff
@@ -284,6 +288,19 @@ def get(self, request):
         return Response(response)
 
 
+class CheckView(generics.GenericAPIView):
+    def get(self, request):
+        response = helpers.search_commodity_by_term(term='feta', json=False)
```
bobmeredith, line 293: I think this should maybe go inside the try catch, I think it would throw on a non 200 response.

core/helpers.py
```diff
@@ -235,7 +235,10 @@ def search_commodity_by_term(term):
     )
 
     response.raise_for_status()
-    return response.json()
+    if json:
+        return response.json()
+    else:
+        return response
```
bobmeredith, line 241: This line doesn't get coverage. Could refactor to a `return response.json() if json else response` to get round that :)

agiamas: hah, a 1 liner in Python, how Rubyist =) 😂

tests/unit/core/test_views.py
agiamas: checking this with the test_http_error.

bobmeredith (approved):
LGTM

PR #651, opened 2020-11-10
use compile css, replace ditMVP.js to magna.js and ditMVP.css to styl…

dahfool:
use compile css, replace ditMVP.js to magna.js and ditMVP.css to styles.css


lewis-coulson-dit (approved):
LGTM

PR #657, opened 2020-11-11
GP2-652: Handle expired staff SSO token that causes a 500 on Dev, Staging, Beta when Staff SSO token has expired 

stevejalim:
This should, I hope, stop the issue where Wagtail 500s when faced with an expired SSO session by re-triggering the SSO auth flow.

It makes sure we only send people to the relevant admin login (Wagtail or Django) if we know were trying to go there.

(It MAY be we need to also zap the cookies, if this isn't enough.)


sso/middleware.py
```diff
@@ -1,11 +1,35 @@
...
-class AuthenticationMiddleware(django.contrib.auth.middleware.AuthenticationMiddleware):
+class AuthenticationMiddleware(DjangoAuthenticationMiddleware):
 
     def process_request(self, request):
-        user = auth.authenticate(request)
+        try:
+            user = auth.authenticate(request)
+        except TokenExpiredError:
+            resolver_match = resolve(request.path)
+            if (
```
depsiatwal, line 20: why the () around if 
since we don't use this syntax found it a little hard to read at first glance

sso/middleware.py
```diff
@@ -1,11 +1,35 @@
...
+                # Covers Django admin
+                resolver_match and resolver_match.namespace == 'admin'
+            ):
+                return HttpResponseRedirect(reverse('admin:index'))
+            elif (
+                # Covers our use of Wagtail admin
+                request.path_info.startswith('/admin/')
+            ):
+                return HttpResponseRedirect(settings.LOGIN_URL)
```
depsiatwal, line 29: may not following but should this not be the other way round
Wagtail goes to wagtail login fall back is to user login screen

sso/middleware.py
stevejalim: Just to put the comment with the condition in a way that doesn't make the line too long. It's similar to how Black auto-formats long lines with comments at the end

sso/middleware.py
stevejalim: `settings.LOGIN_URL` is the Wagtail login, either authbroker/Staff SSO (if a deployed version of the codebase)or username+password (if local)

depsiatwal (approved):
lgtm

PR #669, opened 2020-11-13
GP2-741: Bump Wagtail Transfer to latest version + update config

stevejalim:
This changeset moves us to 0.6, which supports copying objects that use multi-table inheritance, and also handles copying tags on items.

It also reduces the scope of the configuration so that no sites can pull content from Beta, because - for now - we're going to get WT in use up to staging, and then see what extra features we need to support before we can use it to suit our workflows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,7 @@
 
 ### Implemented enhancements
 
+- GP2-741 - Update Wagtail-Transfer version and configuration
 - GP2-914 - Responsive header
 - GP2-978 - Mobile learning homepage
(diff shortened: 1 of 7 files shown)


webbyfox (approved):
Wondering how we cover these config changes in the tests.

stevejalim:
> Wondering how we cover these config changes in the tests.

I think it's hard to prove the config is correct with unit tests - just manual testing. However, I could add a comparison of current settings with settings we (now) know work, to protect against regressions. Plus some static checks to ensure any new models that should be featured in the config are not missing... what do you think?

PR #673, opened 2020-11-16
GP2-984 Compare market selection and redux global state management

bobmeredith:
This ticket adds a selection button block at the bottom of the 'compare markets' page. 
The big thing about this is that we want to reflect changes in the markets-under-comparison, the list of buttons to select from and the current selected product in the personalisation bar, all together.

To do this, we need global scope, so I have restored the redux store which was already implemented and fixed its initialization. Additionally, this opens up scope for removing many of the properties that we currently connect to React components at the insertion site, and moving them into initialized global scope.

Also, we can manage persistence through the redux reducers. I have used this to persist the market in export plan but this required a considerable re-factor of the Services.js module to avoid cirular dependencies.  I've moved the config section into its own module (config.js) and all the api calls (api.js).  Services is only responsible for creating the store but passes through the api and config to minimize other code changes

ATM, all the reducers are in one file but could easily be split out as they grow.  I've left the old ones in as some of the existing redux state was used for the sign-up modal.

 - [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-984

 - [ ] (if needs manual testing) How to test locally, including how to set up appropriate data


bobmeredith:
image

CHANGELOG.md
```diff
@@ -136,6 +136,7 @@
 - GP2-315 - route to market backend integration
 
 ### Implemented enhancements
+- GP2-984 - Compare markets selector and reducx global state managememt
```
lewis-coulson-dit, line 139: reducx, don't tell me that's another JS library :)

core/templates/core/target_markets.html
```diff
@@ -28,30 +32,27 @@
...
+    </div>
+  {% endif %}
+{% endblock %}
 
-  </div>
+{% block additional_content %}
+  {% if data_tabs_enabled %}
+    <div id="open-product-finder" class="m-v-l" data-productname="{{export_plan.export_commodity_codes.0.commodity_name}}" data-productcode="{{export_plan.export_commodity_codes.0.commodity_code}}"></div>
+    <script>
+      magna.CompareMarkets({
```
lewis-coulson-dit, line 43: Could you not put this initialization code in the script tag below?

react-components/src/Services.js
```diff
@@ -1,298 +1,35 @@
...
-    method: 'get',
-    headers: {
-      Accept: 'application/json',
-      'X-CSRFToken': config.csrfToken,
-      'X-Requested-With': 'XMLHttpRequest'
-    }
-  })
-}
-
+/* Unused
```
dahfool, line 14: I suggest we just delete what is unsed.

CHANGELOG.md
bobmeredith: Hehe

core/templates/core/target_markets.html
bobmeredith: Yes, good point - no point in two script tags.
Will change

react-components/src/api.js
```diff
@@ -0,0 +1,175 @@
+/* eslint-disable */
+import { config } from '@src/config'
+
+const MESSAGE_UNEXPECTED_ERROR = { __all__: ['Unexpected Error'] }
```
dahfool, line 4: Might to be good to move all actions in one place for consistency, reusablity and will help when writing test.

react-components/src/components/CompareMarkets/SelectMarket.jsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
```
dahfool, line 1: No necessarily a problem, seperating the component from the container would help for unit test.

react-components/src/components/Sidebar/CountryNotSelected/index.jsx
```diff
@@ -32,11 +32,11 @@ export const CountryNotSelected = ({ isOpen }) => {
           type={1}
         />
       </ReactModal>
-      <CountryFinderModal
-        modalIsOpen={modalIsOpen}
-        setIsOpen={setIsOpen}
-        addButton={false}
-      />
+      {/*<CountryFinderModal
```
lewis-coulson-dit, line 35: Did you mean to leave this in here?

react-components/src/components/Sidebar/index.jsx
```diff
@@ -14,9 +14,9 @@ export const Sidebar = ({ sections, logo, company, currentSection }) => {
         onClick={() => setModal(false)}
         isOpen={modal}
       />
-      <CountryNotSelected
+      {<CountryNotSelected
```
lewis-coulson-dit, line 17: why the curly braces

react-components/src/components/Sidebar/CountryNotSelected/index.jsx
dahfool, line 35: please delete `CountryFinderModal`

react-components/src/api.js
bobmeredith: Not sure I understand. You mean the redux action definitions in actions.js?
I've moved all the api calls (xhr gets/posts/delete) into this file from services.  This is so that we can call them from reducers because otherwise, there would be a circular module reference. 

Services.js now only contains the store creation (importing the reducers). To limit the amount of code refactoring, Services,js imports config from config.js and api calls from api.js and adds them to its export so the facade remains the same.

dahfool: my bad, i mean the constants .eg `MESSAGE_PERMISSION_DENIED`.

react-components/src/bundle.js
```diff
@@ -41,11 +43,12 @@ export default {
...
   createCostsAndPricing,
   ProductFinderButton,
-  CountryFinder,
+  CountryFinderButton,
   CompareMarkets,
+  SelectComparisonMarket,
   ModalMenu,
   createLogin,
-  setConfig: Services.setConfig,
+  setConfig: setConfig,
```
dahfool, line 51: es6 `setConfig,` should be enough

bobmeredith: Yup, thanks

react-components/src/components/Sidebar/CountryNotSelected/index.jsx
bobmeredith: No - and great spot - I need to fix up  this trigger of the modal now it's been split up.

dahfool:
Nice refactor 🕺🏿

react-components/src/Services.js
bobmeredith: You're braver than me!
will do

react-components/src/components/Sidebar/index.jsx
bobmeredith: Good spot - thanks

react-components/src/api.js
bobmeredith: Makes sense - will do.
Put them in a file contsnts.js.
exports each message plus 'messages' as a map.  That gets made available through services.js too for backward compatibility

react-components/src/components/CompareMarkets/SelectMarket.jsx
bobmeredith: Ah yes, good point.
I've not actually written a 'unit' test for this component, but rather an integration test that covers its interaction with the other components on the page as I think that's more useful.

dahfool: Fair. I am writing unit tests for react component and believe that the end to end test will cover integration.

dahfool (approved):

PR #675, opened 2020-11-16
Feature/gp2 283 direct cost component

dahfool:

 - [x] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-283 


react-components/src/components/CostsAndPricing/Units/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+      <h2 className='h-xs p-t-l p-b-0'>Total costs and price</h2>
+      <p>Now you have calculated your direct and overhead costs, you can calculate your final cost per unit. This can be
+        tricky but don't worry, we will tell you what you need to do.</p>
+
+      <h2 className='h-xs p-t-l p-b-0'>Number of units you want to export</h2>
+      <p>The first thing you can do is record how many units of your product you want to export in the following
+        tool.</p>
+      <p>Try to be as accurate as possible, this will help make your plan stronger.</p>
+      <div className='c-1-6 m-r-xs'>
+        <Input />
```
bobmeredith, line 17: Doesn't this need some props? Id, label and update at least.

react-components/src/components/CostsAndPricing/Units/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+      <div className='c-1-3'>
+        <Select label='' update={() => {}} name='test' options={[]} />
```
bobmeredith, line 20: Likewise - I guess the update fn yet to be built as the endpoint not there yet?

react-components/src/components/Form/FormGroup/index.jsx
```diff
@@ -24,7 +24,7 @@ export const FormGroup = ({
         type='LEFT'
       />
     }
-    <label className='form-label' htmlFor={id}>{label}</label>
+    { label && <label className='form-label' htmlFor={id}>{label}</label> }
```
bobmeredith, line 27: Don't you break accessibility by leaving a label out?

react-components/src/components/Form/FormGroup/index.jsx
```diff
@@ -45,7 +45,7 @@ FormGroup.propTypes = {
   errors: PropTypes.arrayOf(PropTypes.string),
   tooltip: PropTypes.string,
   id: PropTypes.string.isRequired,
-  label: PropTypes.string.isRequired,
```
bobmeredith, line 48: As above

bobmeredith (commented):
Not sure about making labels optional as it breaks accessibility I think.
If the design doesn't call for visible labels - maybe pass a flag to hide the label as well as the label text - then add a 'visually-hidden' class  to it

react-components/src/components/CostsAndPricing/Units/index.jsx
dahfool: we have a couple of designs of input field without label.

react-components/src/components/CostsAndPricing/Units/index.jsx
dahfool: No logic have been update to these component yet, purely UI for styling.

react-components/src/components/Form/FormGroup/index.jsx
dahfool: Will update to hide, rather than removing.

bobmeredith (approved):
LGTM

PR #676, opened 2020-11-16
Re-adding the DIT link that was removed by another PR

jakehopking:

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-932
 

depsiatwal (commented):
not sure we should be hard coding urls in templates 
I would drive this from BE and add via Env variable 
we can add ENV variables read settings using url template tag

depsiatwal (approved):
LGTM

PR #679, opened 2020-11-16
Move analytics code inside promise 

webbyfox:


react-components/src/components/RouteToMarket/index.jsx
```diff
@@ -37,27 +37,22 @@ export const RouteToMarket = ({
       .catch(() => {})
   }
 
-    const pushAnalytics = () => {
```
lewis-coulson-dit, line 40: I'd be tempted to abstract this out into a hook so we don't have to repeat this code everywhere we want to use analytics. Something like this should do the trick:

`// useAnalytics hook
function useAnalytics() {
  const [isAnalyticsCalled, setIsAnalyticsCalled] = useState(false);

  analytics(obj)
  setIsAnalyticsCalled(true)

  return {
      analytics: (obj) => {
        analytics(obj)
        setIsAnalyticsCalled(true)
      },
      isAnalyticsCalled
  };
}


// consume it like this
const [analytics, isAnalyticsCalled] = useAnalytics()


// call it like this
if (isAnalyticsCalled) {
    analytics({
        'event': 'planSectionSaved',
        'sectionTitle': 'route-to-market'
    })
}
`

lewis-coulson-dit (approved):

PR #683, opened 2020-11-17
GP2-849-save-progress

depsiatwal:


exportplan/templates/exportplan/sections/marketing-approach.html
```diff
@@ -86,7 +87,10 @@ <h2 class="h-m m-b-xs">How will you get your product to customers in {{ export_p
           {key: "55-64", label: '55-64 year olds'},
           {key: "65+", label: '65 years old and over'}
       ],
-      country: '{{ export_plan.export_countries.0.country_name }}'
+      country: '{{ export_plan.export_countries.0.country_name }}',
+      insight: {{ demographic_data|safe }},
```
bobmeredith, line 91: somehow, this feels a bit wrong - constructing JSON in a an HTML file with un-escaped fields - particularly as part of the structure uses a json stringifier.  How about constructing this object as a dict in python in the view, then using just `magna.createTargetAgeGroupInsights(element, {{ageGroupInsightsParameters|to_json}}`
You'd need to split out the element from the other things of course.

exportplan/templates/exportplan/sections/target-markets-research.html
```diff
@@ -78,7 +79,9 @@ <h2 class="h-xs p-t-l p-b-0">Data Snapshot: {{ export_plan.export_countries.0.co
           {key: "65+", label: '65 years old and over'}
       ],
       country: '{{ export_plan.export_countries.0.country_name }}',
-      insight: {{ insight_data|safe }}
+      insight: {{ insight_data|safe }},
+      selected: {{ selected_age_groups|safe}},
+      currentSection: {{ current_section|to_json}},
     })
```
bobmeredith, line 85: Same as above comment.  In fact, if the parameters are both the same, even more reason to build a common object on the python side.

exportplan/templates/exportplan/sections/marketing-approach.html
depsiatwal: good point 
I don't think this is need any more will update

react-components/src/components/TargetAgeGroupInsights/AgeGroupData/index.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { memo } from 'react'
+import { Stats } from '@src/components/Stats'
+import { notAvailable, StatsGroup } from '@src/components/Stats/StatsGroup'
+import PropTypes from 'prop-types'
+
+export const AgeGroupData = memo(
+  ({ targetPopulation, female, male, urban, rural }) => (
+    <>
+      <h3 className="body-l-b m-t-s m-b-xs">
+        Data specific for you target age group
```
bobmeredith, line 10: Looks like a typo (your?)

bobmeredith (approved):
LGTM

PR #686, opened 2020-11-18
GP2-661: Support S3-to-S3 copying for Wagtail Transfer

stevejalim:
Takes the approach of a custom FileAdapter that is only used if all assets are on S3.

Key points:

* Uses the hash of the file from the S3 ETag rather than load and hash the entire file - saves time. See the comments re the use of the ETag as a hash: there are some limitations.
* Uses S3's client.copy() API to do the copying legwork, rather than "download, stream accross, upload" - saves time
* ONLY works if the source file is publicly readable
* ONLY works within the same AWS Region (according to the docs - untested)

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-661 
 
### DIFF
--- a/tests/unit/core/test_wagtail_hooks.py
+++ b/tests/unit/core/test_wagtail_hooks.py
@@ -4,11 +4,22 @@ from unittest import mock
 from datetime import timedelta
 
 import pytest
+
+from boto3.exceptions import RetriesExceededError, S3UploadFailedError
+
+from django.db.models import FileField
+from django.test import override_settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.sessions.middleware import SessionMiddleware
 from wagtail.core.rich_text import RichText
 
 from core import wagtail_hooks
+from core.wagtail_hooks import (
+    register_s3_media_file_adapter,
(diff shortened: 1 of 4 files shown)


stevejalim:
Deleted the old codecov report - follow the link in the checks to see it

tests/unit/core/test_wagtail_hooks.py
```diff
@@ -553,3 +564,548 @@ def test_case_study_modeladmin_list_display_methods():
...
+    mock_S3WagtailTransferFile = mock.Mock(return_value=mock_s3_file)  # noqa N806
+
+    with mock.patch(
+        'core.wagtail_hooks.S3FileFieldAdapter._get_relevant_s3_meta',
+        mock_get_relevant_s3_meta
+    ):
+        with mock.patch(
+            'core.wagtail_hooks.S3WagtailTransferFile',
+            mock_S3WagtailTransferFile
+        ):
```
webbyfox, line 834: Above two patches can be done in one `with` statement

webbyfox (approved):
Looks good... Just minor comment but happy to approve :+1:

PR #695, opened 2020-11-23
Merging release -> master to keep it up to date

stevejalim:
Keeping `master` up to date with _almost_ everything that is going to Beta/UAT shortly.

https://github.com/uktrade/great-cms/tree/beta-EP-release-stablity is based on `release` but has a couple of fixes which do NOT need to be backported to `develop`, because they already exist there. As such, `develop` will feed into `release` and `release` will eventually feed into `master` so the single commit that's different on that branch (bc2770339) does not need to exist in `release` or `master`. (`master` will effectively 'catch up' with `beta-EP-release-stablity`)

Note that the result of this merge will NOT be deployed to UAT/Beta - but https://github.com/uktrade/great-cms/tree/beta-EP-release-stablity will be deployed to UAT/Beta instead


stevejalim:
Am not moving ahead with this merge - there's a regression on `release` due to a breaking change in `great-styles` that means we can't rebuild assets right now. The fix for the breaking change is on `develop` and will become part of `release` tomorrow.

We didn't need to merge this branch to be able to release to UAT, because there was a fork of `release` made BEFORE the breaking change was done, which was handy. The use of that branch is NOT a formal release.

The downside of NOT merging this now is that the next time we do a cut to master from release, it will have two sprints' worth of stuff in it, not one.

cc @webbyfox @depsiatwal @agiamas @bobmeredith  @jakehopking @dahfool @rafa-garcia for awareness. Can you give me an emoji reaction, please, so I know I'm not just shouting at clouds? :)

agiamas:
👍👍😎🙌

dahfool:
💥

PR #700, opened 2020-11-24
Reverting back to coming soon for Target Market page

webbyfox:


config/env/dev
```diff
@@ -38,6 +38,7 @@ STAFF_SSO_AUTHBROKER_URL=https://www.examole.com
 AUTHBROKER_CLIENT_ID=debug
 AUTHBROKER_CLIENT_SECRET=debug
 FEATURE_FLAG_HARD_CODE_USER_INDUSTRIES_EXPERTISE=True
+FEATURE_COMPARE_MARKETS_TABS={"population":true}
```
stevejalim, line 41: Is this also set in Vault yet for stage and beta? I see it has a safe default anyway

config/settings.py
```diff
@@ -509,6 +509,7 @@ def get_client_ip(request):
 FEATURE_FLAG_HARD_CODE_USER_INDUSTRIES_EXPERTISE = env.str('FEATURE_FLAG_HARD_CODE_USER_INDUSTRIES_EXPERTISE', False)
 FEATURE_EXPORT_PLAN_SECTIONS_DISABLED = env.str('FEATURE_EXPORT_PLAN_SECTIONS_DISABLED', False)
 FEATURE_ENABLE_PRODUCT_SEARCH_WHEN_NO_USER = env.bool('FEATURE_ENABLE_PRODUCT_SEARCH_WHEN_NO_USER', False)
+FEATURE_COMPARE_MARKETS_TABS = env.str('FEATURE_COMPARE_MARKETS_TABS', '{ }')
```
stevejalim, line 512: django-environ gives us `env.dict` which would mean we could skip `json.loads()` but we can switch to using that in the tech-debt sprint - no need to change now if you've got a deadline

core/templates/core/target_markets.html
```diff
@@ -12,29 +14,55 @@
...
-  <div class="p-v-l"></div>
+  <div class="hidden" id="next-steps">
+      {% for component in dashboard_components  %}
+        {% if component.value.route_type == 'learn' or component.value.route_type == 'plan' %}
+          <article class="dashboard__routing-block c-1-2 wide">
+            {% include_block component with wide=True %}
+          </article>
+        {% endif %}
+      {% endfor %}
+  </div>
```
stevejalim, line 51: I am not sure this next-steps block should be here -- I thought it was only due to go from develop to staging today

stevejalim (approved):
Approved, but with a key question about the next-steps block -- please check before merging

config/env/dev
webbyfox: in vault for staging and beta... it should be empty dict

core/templates/core/target_markets.html
webbyfox: I think it still will be disable so should be ok

PR #709, opened 2020-11-27
GP2-1144: Allow Beta token checks to support sub-paths of allowed paths

stevejalim:
This changeset should reduce the noise in Sentry for paths which lack a Beta token but don't need the token anyway.

Plus, this fixes a broken UI issue on Beta:  `/static/` is currently configured as an allowed path, but only as a distinct, _complete_ path, rather than the head of a longer path to some static assets. This means that the login/landing page for a user on Beta who does NOT have a token lacks all the JS and CSS we'd want them to see. This bug should be fixed by these changes.

Adds specific tests, and not just to keep coverage levels up ;)


core/middleware.py
```diff
@@ -99,10 +99,16 @@ def __init__(self, get_response):
...
         response = self.get_response(request)
-        # need to whitelist the endpoint, to be able to generate tokens
-        # == '/api/create-token/' or request.path == '/favicon.ico':
-        if request.path in settings.BETA_WHITELISTED_ENDPOINTS:
-            return response
+        # Need to whitelist the endpoint, to be able to generate tokens
+        # == '/api/create-token/' or request.path == '/favicon.ico'.
+        # Note that we now support sub-paths of the whitelisted endpoints,
+        # not just the absolute endpoint paths
+        for whitelisted_path in settings.BETA_WHITELISTED_ENDPOINTS.split(','):
```
stevejalim, line 106: Note the change here from substring checking to actually splitting the comma-separated list

stevejalim: Also, there is a risk, I think, of crafted URLs (eg starting with an allowed path) to bypass the token check eg `/favicon.ico/extra/path/here/` but I don't think this is likely to reveal much info as it'll just 404. 

Do reviewers agree?

agiamas: yes

agiamas (approved):
lgtm

core/middleware.py
```diff
@@ -99,11 +99,20 @@ def __init__(self, get_response):
     def __call__(self, request):
 
         response = self.get_response(request)
-        # need to whitelist the endpoint, to be able to generate tokens
-        # == '/api/create-token/' or request.path == '/favicon.ico':
-        if request.path in settings.BETA_WHITELISTED_ENDPOINTS:
+
+        # Always allow the homepage to be accessed without a token
+        if request.path == '/':
             return response
```
stevejalim, line 105: @agiamas Updated to explicitly always allow `/` but not subpages of it (unless they are in the allowed list, of course)

agiamas (approved):
lgtm

config/env/test
```diff
@@ -21,5 +21,5 @@ AIRTABLE_API_KEY=debug
 DIT_HELPDESK_URL=http://www.debug.com
 FEATURE_FLAG_HARD_CODE_USER_INDUSTRIES_EXPERTISE=false
 BETA_TOKEN=z7445AQ6aGXoPdFxYXQaJPLa3XU6I0zZUuI-Kbar8Jk=
-BETA_WHITELISTED_ENDPOINTS='/api/create-token/,/favicon.ico,/api/check/,/admin/,/auth/login/,/'
+BETA_WHITELISTED_ENDPOINTS='/api/create-token/,/favicon.ico,/api/check/,/admin/,/auth/login/'
```
stevejalim, line 24: Fixed here and in local dev conf

agiamas (approved):
lgtm

PR #723, opened 2020-12-07
Beta Release

webbyfox:


stevejalim:
Just a note (which Riz already knows) - because the last Beta release was using a fork off `release` that included some fixes that did not need to be backported, _this_ diff between `release`  and `master` is twice the size of usual, because half of the changes are already on Beta, via that fork off `release`

PR #726, opened 2020-12-07
Post release changelog update

webbyfox:
Also, backported hotfix from release to develop


stevejalim (approved):
LGTM assuming the JS deps are ok (which I trust you on!)

PR #768, opened 2020-12-21
Dynamic title tags on pages

rafa-garcia:
This PR enables adding title tags with dynamic titles to both Wagtail and Django pages.

### Workflow

 - [X] Ticket exists in Jira  https://uktrade.atlassian.net/browse/GP2-1193 

### Merging


core/mixins.py
```diff
@@ -84,3 +84,15 @@ def add_ga360_data_to_payload(self, request):
         self.ga360_payload['login_status'] = is_logged_in
         self.ga360_payload['user_id'] = user.hashed_uuid if is_logged_in else None
         self.ga360_payload['site_language'] = translation.get_language()
+
+
+class PageTitleMixin(object):
+    # used by views to set a page_title attribute
+    def get_page_title(self, context):
+        return getattr(self, 'page_title', 'Welcome to great.gov.uk')
```
bobmeredith, line 92: This default is not needed as it's defined in the base template

core/views.py
```diff
@@ -71,9 +72,10 @@ def __init__(self):
         )
 
     template_name = 'core/login.html'
+    page_title = 'Log in | great.gov.uk'
```
bobmeredith, line 75: Shouldn't the ' | great.gov.uk' part apply to all pages?

bobmeredith (commented):
First, I think the '| great.gov.uk' part of the title should be defined in the base class rather than duplicated in the views.  Then it can be applied to the wagtail pages as well.
Second, consider using block overrides in the page templates to set the title, rather than literals in Python.  The base view would of course still define the 'page.title' and '| great.gov.uk' parts.
Lastly, if you do want to define the titles in python, consider just adding context['page' = self. Then, any fields of the view will be visible to templates automatically - so 'page.title' would just work if you set a field called 'title' in the view.

It's sad we don't have any base classes for views defined rather than putting everything into an ever increasing list of mixins.  This seems a rather clunky way to do everything IMHO.

core/mixins.py
```diff
@@ -84,3 +84,15 @@ def add_ga360_data_to_payload(self, request):
...
+
+class PageTitleMixin:
+    # used by views to set a page title attribute
+    def get_page_title(self):
+        return getattr(self, 'title')
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.setdefault('page', {})['title'] = self.get_page_title()
+
```
bobmeredith, line 97: I was thinking simpler than that.  Just `context[page] = self` will do the job nicely.  get_page_title can go completely.

rafa-garcia: But that way `page` would be an exclusive domain of `PageTitleMixin`. I'm thinking that this mixin should only own the attribute `title` in the context of `page`

bobmeredith: OK - I can't find a way to access the view object directly which seems a bit rubbish.  I can see this approach is attempting to prevent future conflicts.

bobmeredith (approved):
LGTM

PR #802, opened 2021-01-15
GP2 1284: Port market guide page from BAU (with necessary scaffolding/CSS etc)

stevejalim:
This changeset is huge. We're sorry about that. This is because it:
* ports over the first page (Wagtail model + HTML + supporting CSS/JS) from BAU into Magna
* needed to bring the base templates and required styling from `directory-components` and `great-domestic-ui` 

So, the result is:
* You can add a `CountryGuidePage` as a child of the domestic homepage, and populate it with all the same data that is used in BAU. 
* The `CountryGuidePage` model from BAU was brought across and stripped down or refactored to only contain relevant code, so we drop all the code related to serialization for the headless-CMS approach, etc. It's a lot simpler, but there may still be some additional logic to add to this page after feedback.
* Most of the CMS Admin page for `CountryGuidePage` looks the same, but a few areas with lots of repeated fields being turned into streamfields
* The header and footer links use hard-coded URLS for now, but these will be updated in future sprints

### Workflow

### Reviewing help

  - Pull the branch and run migrations
  - Add CountryGuidePage as a child of the Domestic homepage
  - populate with data, including the stats accordions etc -- check the validation is working, etc

greatcms trade great_8020_admin_pages_129_edit_

greatcms trade great_8020_united-states_

### Housekeeping

- [ ] ~Added all new environment variables to Vault.~
- [ ] ~Cleaned up old feature flags~
- [ ] ~Upgraded any vulnerable dependencies.~
- [ ] ~I have updated security dependencies.~
- [ ] ~Python requirements have been re-compiled.~

### Merging

- PLEASE LET LEWIS OR STEVE MERGE THIS so that we time it right - we don't want this in the next staging release as it needs more QA time in dev first


.prettierignore
```diff
@@ -0,0 +1,2 @@
+# Ignore all HTML files:
+*.html
```
stevejalim, line 2: This was because prettier was auto-formatting Django templates and breaking tags with `{%` in them

config/env/dev
```diff
@@ -28,6 +28,7 @@ DIRECTORY_FORMS_API_BASE_URL=http://forms.trade.great:8011
 DIRECTORY_FORMS_DEFAULT_TIMEOUT=30
 DIRECTORY_API_CLIENT_BASE_URL=http://api.trade.great:8000
 DIRECTORY_API_CLIENT_API_KEY=debug
+ENVIRONMENT_CSS_THEME_FILE=core/css/environment_dev_theme.css
```
stevejalim, line 31: This copies the location that BAU has, so it should "just work" for everyone

config/settings.py
```diff
@@ -62,6 +62,7 @@
     'core.apps.CoreConfig',
     'cms_extras.apps.CmsExtrasConfig',
     'domestic',
+    'international',
```
stevejalim, line 65: Added an empty app because we needed to put some FE code for international somewhere, when taking it from directory-components

core/blocks.py
```diff
@@ -261,3 +262,70 @@ def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context=parent_context)
         context = self._annotate_with_case_study(context)
         return context
+
```
stevejalim, line 265: These blocks are all new, as part of using a streamfield instead of multiple sets of repeated fields

core/fields.py
```diff
@@ -10,3 +11,8 @@ def single_struct_block_stream_field_factory(
         **kwargs,
     )
     return field
+
+
+# Do not delete - needed for a temporary legacy migration dependency
```
stevejalim, line 16: This is here because initially i copy-pasted the entire model in a lift-and-shift way but then decided a rebuild/refactor was better. However, by that point a migration already referenced the MarkdownField, so I couldn't completely remove it. 

We can drop it once this code reaches prod/beta though, and remove the import in the relevant migration

core/migrations/0049_data_bootstrap_tag_and_industrytag_snippets.py
```diff
@@ -0,0 +1,83 @@
+# Generated by Django 2.2.14 on 2021-01-11 16:55
```
stevejalim, line 1: I noticed some Snippets needed data adding to work, so this is for that

domestic/models.py
```diff
@@ -1,18 +1,55 @@
 from django.conf import settings
```
stevejalim, line 1: This file has the most Python additions, because it's where the Wagtail model for CountryGuidePage was added

domestic/models.py
```diff
@@ -1,18 +1,55 @@
...
+class BaseContentPage(Page):
+    """Minimal abstract base class for pages ported from the V1 Great.gov.uk site"""
+
+    promote_panels = []  #  Hide the Promote panel
+    folder_page = False  # Some page classes will have this set to true to exclude them from breadcrumbs
+
+    class Meta:
+        abstract = True
+
+    def get_ancestors_in_app(self):
```
stevejalim, line 30: Tests for this (and for get_breadcrumbs) can only come when we have ported enough pages to make a realistic page tree to test. I've added skip-marked tests so we remember to add them later

config/settings.py
webbyfox, line 65: do we need `international` app atm?  I see an empty app.

webbyfox (commented):
Thanks, @stevejalim  @lewis-coulson-dit  Added some comments for clarification.  I would invite @bobmeredith  or @dahfool to look have look at FE stuffs.

domestic/models.py
stevejalim: Yes, but at the moment, CountryGuidePage can only have one ancestor, which itself can only have the root as it ancestor. Next ticket we will be adding the ArticlePage properly, giving us a tree depth of three and a more realistic set of models to test with

core/components/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+To do (delete all that do not apply):
+
+ - [ ] Changelog entry added.
+ - [ ] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
+ - [ ] (if updating requirements) Requirements have been compiled.
+ - [ ] (if adding CSS) CSS have been compiled.
+ - [ ] (if adding HTML) Change is accessible to the standards we subscribe to.
+ - [ ] (if adding env vars) Added any new environment variable to vault.
+ - [ ] (if adding feature flags) Cleaned up old flags
+ - [ ] (if changing the UI) Add a printscreen to the PR
```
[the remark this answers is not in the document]
stevejalim: I've deleted it while you were reviewing - thanks for the spot!

core/templates/components/google_tag_manager_body.html
```diff
@@ -0,0 +1,4 @@
+<!-- Google Tag Manager (noscript) -->
+<noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ directory_components_analytics.GOOGLE_TAG_MANAGER_ID }}{{ directory_components_analytics.GOOGLE_TAG_MANAGER_ENV|safe }}"
+height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+<!-- End Google Tag Manager (noscript) -->
```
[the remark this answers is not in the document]
stevejalim: Right now we're just moving the Magna tracking over, but we will have to settle on a consistent way in a future ticket

core/templates/components/cookie_notice.html
```diff
@@ -0,0 +1,24 @@
...
+        </div>
+      </div>
+      <div class="cookie-confirmation">
+        <div class="column-full">
+          <p>{% trans "You've accepted all cookies.  You can "%}<a class="link" href="{{ header_footer_urls.cookie_preference_settings }}">{% trans 'change your settings' %}</a> {% trans 'at any time.' %}</p>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
```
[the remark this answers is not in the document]
stevejalim: That's a great question and I'll add it to the list of things to sort out!

config/settings.py
stevejalim: It was there for Lewis to add static assets, etc to when porting the required parts of `directory-components` -- @lewis-coulson-dit do we need `internatonal` yet after all? If not i'll remove it

lewis-coulson-dit: Can remove this for now. Will need to add when we start rebuilding international pages.

stevejalim: Removing now

domestic/models.py
stevejalim: (The tests for this will 1000% come, I promise you)

jakehopking (approved):
Essentially impossible to review, however from what we've discussed: LGTM 🚀

PR #833, opened 2021-01-22
GP2-1363: Drop custom Django view serving /markets/

stevejalim:
This changeset removes the Django view serving /markets/ because this URL path is required for pages that will be ported from Great V1.

The view is not being moved to respond to a different route - it is no longer required, so it is safe to remove.

Related tests were removed.

Other tests that used /markets/ as an example URL have been updated so that they still pass

### Workflow

### Merging


core/urls.py
```diff
@@ -20,7 +20,6 @@ def anonymous_user_required(function):
 
 
 urlpatterns = [
-    path('markets/', views.MarketsView.as_view(), name='markets'),
```
webbyfox, line 23: dont we need to get rid of this MarketView as well?

stevejalim: Yep - janky keyboard here meant the edits on views.py didn't get saved. Updating now

core/views.py
```diff
@@ -88,31 +88,6 @@ def __init__(self):
...
-class MarketsView(GA360Mixin, TemplateView):
-    def __init__(self):
-        super().__init__()
-        self.set_ga360_payload(
-            page_id='Markets',
-            business_unit='MarketsUnit',
-            site_section='markets',
-        )
-
-    template_name = 'core/markets.html'
```
webbyfox, line 100: Does this template need to remove as well, then?

stevejalim: Yep - thanks! 🤦

webbyfox (approved):
LGTM

PR #844, opened 2021-01-25
GP2-1368: Wire up link from country_guide.html to /advice/ slug

stevejalim:
Before it was trying to reverse URL which didn't exist in Django

I know that using `slugurl` isn't much different from using a hard-coded URL, that it still depends on a fixed value _somewhere_ in the system, but Advice is a section we know will 100% exist, and we have to make the link somewhere, so here is at least clearer than using context processors, etc, now that we no longer need to.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1368

### Merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Implemented enhancements
 
+- GP2-1368 - Wire up link from country_guide.html to /advice/ slug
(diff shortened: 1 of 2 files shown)


webbyfox (approved):
LGTM

PR #862, opened 2021-01-27
Gp2-1382-getting-paid-structure-integration

depsiatwal:
### Workflow

### Merging


lewis-coulson-dit (approved):
FE LGTM

exportplan/views.py
```diff
@@ -291,6 +291,24 @@ def get_context_data(self, *args, **kwargs):
...
+        context = super().get_context_data(*args, **kwargs)
+        context['payment_method_choices'] = choices_to_key_value(choices.PAYMENT_METHOD_OPTIONS)
+        context['payment_term_choices'] = choices_to_key_value(choices.PAYMENT_TERM_OPTIONS)
+        transport_choices = {
+            'All forms of transport': choices_to_key_value(choices.TRANSPORT_OPTIONS),
+            'Water transport': choices_to_key_value(choices.WATER_TRANSPORT_OPTIONS),
+        }
+        context['transport_choices'] = transport_choices
+
+        context['getting_paid_data'] = json.dumps(self.export_plan['getting_paid'])
```
bobmeredith, line 307: I'm wondering about the method for getting json into HTML.  Why is this field being json.dumped whereas other fields like 'transport_choices' don't seem to need it?  They are both being rendered rendered by {{transport_choices|safe}} after all.
I suspect the reason for the json.dumps here is because the default to_string of the dict renders as python to_string values rather than JSON, so true would appear as :True rather than :true.
It may be better to leave the json.dumps out of the view and add the filter {{ xxx|to_json}} to do proper json conversion in the template.

depsiatwal: correct getting_paid here is a dict from database which has python stuff in it. 
FYI i'm refactoring currently so that request.user.exportplan returns a exportplan parser idea is that we can use this to make the EP parse for UI

bobmeredith: AFAICT, choices_to_key_value for instance, is also a python object -  list in that case. I'm thinking there is danger in rendering in the template without the |to_json filter. I mean |safe is anything but!

depsiatwal: true but choices are built using choices_to_key_value which is a util mainly for UI. 
I think also refactoring these out of helper into a UI Utile will help and make it less confusing.

bobmeredith (approved):
LGTM

PR #864, opened 2021-01-27
GP2-1179 add category support to dropdown

dahfool:
Add category support to dropdown

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse//GP2-1179


dahfool:
image

react-components/src/components/Form/Select/index.jsx
```diff
@@ -96,7 +116,6 @@ export const Select = ({
           isOpen ? 'select__button--close' : ''
         }`}
         type="button"
-        role="button"
```
bobmeredith, line 99: You're removing this role as it's already a button so the role is redundant?

bobmeredith (approved):
LGTM - just one tiny question

react-components/src/components/Form/Select/index.jsx
dahfool: > You're removing this role as it's already a button so the role is redundant?

Yes redundant, i was gettingt this warning from eslint - The element button has an implicit role of button. Defining this explicitly is redundant and should be avoided.(jsx-a11y/no-redundant-roles)

PR #888, opened 2021-02-01
GP2-1398 Where-to-export - split data table loading

bobmeredith:
The datatable in 'Where to export' suffers from poor loading performance. In particular, some data are very moch slower than others to retrieve. Currently, all the data for a particular tab are retrieved in a single request to the BE, and the requests are fired out from this view, nose-to-tail, so it's quite possible the whole request will time out. Additionally, all requests from magna BE to directory-api, only request data for a single country.
This PR is the first step in addressing this. The FE data table can now be configured to request data for cells in groups of columns, and each group can be configured to retrieve data either for all countries in one request, consecutively or concurrently.
As part of this, I have split the country data request from the com-trade data request, so that group of columns can be retrieved separately on the economy tab.

Also, split the configuration of the tabs into separate files - one per tab, and rationalised the handling of errors and 'data not available' indicators by pushing more code into the outer renderer.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1398

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
Testing:
ATM, the population and society tabs load in one go.  Add a set of new countries, then observe the table load - all data should populate at once.
The economy tab is different.  The two leftmost columns load separately and one-request per row/country.  The right three columns should be much quicker and load all in one block. You should expect to see the com-trade data (left two columns) load very slowly on new countries. Refresh the page - and you would expect to see the data load much quicker, but still in the same set of operations. All loading indicators should clear (eventually) in all cases.
This update also includes a change to the way 'data not available' is rendered. Check odd countries where there are missing data for certain columns.

rafa-garcia (approved):
LGTM. Tried locally and although it's a huge improvement, the first load of economy data is still painfully slow. Nice effort nevertheless
👍 for splitting up the tab components

PR #897, opened 2021-02-03
GP2-1190: Remove wagtail-personalisation - code-drop 1 of 2

stevejalim:
Currently, `wagtail-personalisation` is blocking our Wagtail upgrade path. 
* https://github.com/wagtail/wagtail-personalisation/pull/212 is a WIP fix for a breaking change in Wagtail, but is stalled at getting the CI tests passing (and I've also tried fixing this myself and it's non-trivial)
* the project doesn't seem to be that actively maintained
* we're not really using it, so we should drop it, especially in case we end up with another upgrade-blocking issue in the future.

However, dropping it is not as easy as you might hope, because there are model bases and streamfield blocks from `wagtail-personalisation`, which means _migrations_ will fail to even load / the server won't start if we simply remove `'wagtail-personalisation'` from settings.INSTALLED_APPS. I've tried a handful of ways to approach this and have settled what looks like a viable plan.

1) [This changeset] - remove all _use_ of wagtail-personalisation from the codebase + add a new migration for that. Get that code deployed all the way to prod/Beta. This is because `wagtail-personalisation` still needs to be 'around' for that new migration to run (because Django has to set up the old DB state before the new one can be migrated to)

2) [Next changeset, next sprint] - squash all migrations, including the one that removed `wagtail-personalisation` then _delete all references to `wagtail-personalisation` from all past migrations_, then remove it from INSTALLED_APPS.

So, this changeset is the first step in that plan. Note that removing wagtail-personalisation's custom `PersonalisedBlock`s does not lose any content - the StreamField handles the change just fine. If we had been using it for segment-targetted content, those that would be made visible to all users, but we don't so it is not a risk 

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1190

### Reviewing help

* To test locally, before pulling the branch, set up two personalised text block in a lesson with content that you can easily identify. Make one for Everyone and another for a Segment (you may need to create a segment first).
* Confirm in the published local page that you can see the Everyone content but not the Segment content
* Pull the branch and run the migration
* Re-visit the lesson page and confirm the content for Everyone is still there and the the personalised segment content is ALSO visible (ie, no rules apply and no content has been lost) 
* Confirm you can edit/save/republish the lesson page
 
BEFORE

Screenshot 2021-02-03 at 10 38 47

AFTER - all content there, just no segment filtering
Screenshot 2021-02-03 at 10 39 48

### Housekeeping

### Merging

- [ ] This PR can be merged by reviewers. **Please leave for Steve to merge after checking in cross-dev**


core/apps.py
```diff
@@ -5,4 +5,4 @@ class CoreConfig(AppConfig):
     name = 'core'
 
     def ready(self):
-        from core import rules  # noqa F401
+        pass
```
webbyfox, line 8: Good to get rid of this function, I suppose

webbyfox (approved):

PR #899, opened 2021-02-03
GP2-1419 Fix ex-ops blow-up on dashboard

bobmeredith:
The dashboard is blowing up on dev for certain users.
It appears that the problem is that the ex-ops request is returning a list of opportunities, but at least one date is 'None'
As it's impossible to reproduce this locally - the simplest fix is just to throw a try-catch around the formatter.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1419

### Reviewing help

### Merging

### DIFF
--- a/core/serializers.py
+++ b/core/serializers.py
@@ -36,7 +36,6 @@ def _date_format(string):
 
 
 def parse_opportunities(results):
-
     return [
         {
             'title': result['title'],
--- a/core/serializers.py
+++ b/core/serializers.py
@@ -29,10 +29,14 @@ class CompanySerializer(serializers.Serializer):
 
 
 def _date_format(string):
-    return datetime.strptime(string[0:10], '%Y-%m-%d').strftime('%d %b %Y')
+    try:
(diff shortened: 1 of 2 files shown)


bobmeredith:
Testing: I cannot reproduce the issue on local or development.
The only reasonable way to test this fix is for known (system) users with the error to check that he problem is fixed.

core/serializers.py
```diff
@@ -29,7 +29,10 @@ def validate_expertise_countries(self, value):
 
 
 def _date_format(string):
-    return datetime.strptime(string[0:10], '%Y-%m-%d').strftime('%d %b %Y')
+    try:
+        return datetime.strptime(string[0:10], '%Y-%m-%d').strftime('%d %b %Y')
+    except Exception:
+        return ''
```
depsiatwal, line 35: do we want all exceptions here or just format exception ?

depsiatwal (approved):
LGTM

bobmeredith:
This change has been included already

PR #903, opened 2021-02-04
GP2-1361 Compare markets - align country list to user

bobmeredith:
We store the list of countries being compared in a cookie. Trouble is, if multiple users log in on the same machine (same user acct on machine) they all get the same set as last saved. This is particularly problematic given that the selected country is in teh export plan and therefore associated with the user.

This is a simple fix to append the user_id to the cookie, so they are unique.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1361

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
Testing.

- Log in as a user.
- go to 'Where to export' and choose one or more countries to compare.
- Log out, then log in as another user on the same browser.
- Go to 'where to export'
- check no countries selected
- select some
- switch user back to the first and check the correct user has the correct set of countries.

Note: any existing countries already selected will be lost by this change.

core/templates/core/base.html
```diff
@@ -70,7 +70,11 @@
           passwordResetUrl: '{{ javascript_components.password_reset_url }}',
           removeCountryDataUrl: '{% url "exportplan:api-remove-country-data" %}',
           termsUrl: '{{ javascript_components.terms_url }}',
-          userIsAuthenticated: {{ request.user.is_authenticated|yesno:"true,false" }},
+          user: {
```
dahfool, line 73: Might be worth just sending the whole {{ request.user}}

bobmeredith: That's a good point.  I tried doing that first, but I get a serialization error - so reverted to this cherry picking.
Remember the user contains the whole export plan too. This could be a nice thing to bring in.  Potentially, it could go into the global state so that all the export plan components can grab from the store.

dahfool (approved):

PR #904, opened 2021-02-04
GP2-1486: Make URL of root breadcrumb configurable

stevejalim:
The breadcrumbs in use on pages ported from V1 feature a hard-coded node as the root breadcrumb, labelled "great.gov.uk"

While we're in the situation where some pages will exist in V2 and others in V1, we need to ensure that the root breadcrumb sends people to V1, not (as was previously the default) the root page of V2/Magna

This will be fine when traffic for certain parts of V1 is channeled to V2, and also when all traffic is going to V2

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1486

### Merging


webbyfox (approved):
LGTM, please add CL

bobmeredith (approved):
LGTM

PR #907, opened 2021-02-05
Draft - GP2-1481: POC for shared header

lewis-coulson-dit:
POC for implementing shared header across Magna and BAU pages.

Approach:

- Styles live in a shared folder in core which are then imported into Magna and BAU CSS builds.
- JS lives in core folder and is included in both Magna and BAU base templates
- Header links will render conditionally based on URL.


webbyfox:
@lewis-coulson-dit  do we need this PR?

PR #912, opened 2021-02-08
GP2-1514: Fix markets topic page ordering

stevejalim:
This changeset fixes two small things with the Markets page ported from V1:
* **Fix ordering of markets page results** 

  Previously, when the markets page was sorted by last updated, it was actually returning pages ordered the least-recently updated because the default ordering of `last_published_at` is ascending not descending.

  This fixes that by amending the relevant selected value before it is passed to the ORM.

  Note that a better querystring name for date ordering would be clearer, but we need to keep the pattern in use on V1 for now

* **Fix verbose title of field not to mention markdown**

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1514

### Merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 - GP2-1411 - Refactor ArticlePage to use a StreamField with new PullQuoteBlock
 
 ### Fixed bugs
-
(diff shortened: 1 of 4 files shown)


webbyfox (approved):

depsiatwal (approved):
LGTM

PR #943, opened 2021-02-17
Amended case study scoring setting name aligning with other fields

webbyfox:
This changeset is fixing case study scoring setting field name aligning with other fields 


stevejalim (approved):
Technically LGTM but also wondering if it's easier to rename the use of the variable in code. Happy either way

webbyfox:
@stevejalim  In Code, field name is dynamically generated

PR #945, opened 2021-02-17
[Hotfix]: Amended case study scoring setting name aligning with other fields

webbyfox:
Amended case study scoring setting name aligning with other fields


stevejalim (approved):
🚢 🚢 🚢

PR #949, opened 2021-02-17
Migration file fix after release hotfix

webbyfox:


stevejalim (approved):
Talked with Riz about this directly.

PR #960, opened 2021-02-18
GP2-1619: Ensure all pages ported from V1 into Domestic support DataLayer

stevejalim:
This changeset makes the inclusion of the two GA360-specific mixins easier
by wrapping them into their own abstract mixin class.

This new `DataLayerMixin` is then added to the `BaseContentPage` for ex-Great-V1 content or
directly into any page that does not use `BaseContentPage`.

There is also a test that checks that all concrete subclasses of wagtail.Page
in domestic.models implement the `DataLayerMixin`. (Opinions welcome about cleaner ways to execute this test.)

eg: without the `DataLayerMixin` included in `BaseContentPage` we get:

```
E           AssertionError: These Domestic pages do not implement the DataLayerMixin, 
but should: [<class 'domestic.models.ArticleListingPage'>, <class 'domestic.models.ArticlePage'>, 
<class 'domestic.models.CampaignPage'>, <class 'domestic.models.CountryGuidePage'>, 
<class 'domestic.models.GreatDomesticHomePage'>, <class 'domestic.models.GuidancePage'>, 
<class 'domestic.models.MarketsTopicLandingPage'>, 
<class 'domestic.models.PerformanceDashboardPage'>, <class 'domestic.models.TopicLandingPage'>]
```

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1619

### Merging

### DIFF
--- a/domestic/models.py
+++ b/domestic/models.py
@@ -31,7 +31,34 @@ from domestic import cms_panels
 from domestic.helpers import build_route_context, get_lesson_completion_status
 
 
-class BaseContentPage(Page):
+class DataLayerMixin(
+    Page,
+    mixins.WagtailGA360Mixin,
+    GA360Mixin,  # from great-components, but could be moved into great-cms
+):
+    """Mixin to automatically set the GA360/DataLayer payload on all pages
+    that implement it"""
+
+    class Meta:
+        abstract = True
+
+    def get_context(self, request):
+        context = super().get_context(request)
+
+        self.set_ga360_payload(  # from GA360Mixin
+            page_id=self.id,
+            business_unit=settings.GA360_BUSINESS_UNIT,
+            site_section=self.slug,
+        )
+        self.add_ga360_data_to_payload(request)
+        context['ga360'] = self.ga360_payload
+        return context
+
+
+class BaseContentPage(
+    DataLayerMixin,
+    Page,
+):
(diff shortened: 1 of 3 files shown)


stevejalim:
Marked draft for a refactor

domestic/models.py
```diff
@@ -31,7 +31,34 @@
...
+    class Meta:
+        abstract = True
+
+    def get_context(self, request):
+        context = super().get_context(request)
+
+        self.set_ga360_payload(  # from GA360Mixin
+            page_id=self.id,
+            business_unit=settings.GA360_BUSINESS_UNIT,
+            site_section=self.slug,
```
stevejalim, line 51: @rafa-garcia @webbyfox @bobmeredith Note that this logic (which was taken from DomesticDashboard, below) does not quite match the logic on `core.CMSGenericPage` where the site section is the URL, not the slug.

I think we should align these two things, but it might break reporting do to so. Any thoughts?

webbyfox: I think that for DM (or PM) to decide but I agreed to align with what we have in Magna. DM can communicate with people who are using these stats maybe?

stevejalim: Tricky thing is we also need to match what V1 does.

lewis-coulson-dit (approved):
LGTM

rafa-garcia (approved):
👍

PR #966, opened 2021-02-22
NOTICKET: Add migraition for missing richtext-features change

stevejalim:
The hotfix to add in the `document-link` option for rich-text blocks worked on master when added to domestic.0016 but migrations on release and develop had moved on (and so the frozen schema no longer reflected the change made to 0016).

This change will remeduy this, but it shows the limitations of editing historical migrations :/

(Editing StreamField schemas is a no-op change at the DB level, so this changeset is 'ok', but it doesn't protect against other edits to the StreamField going on on parallel branches)

Master does not need this fixup, because it doesn't have the extra migrations that staging and develop do, which didn't get the hotfix (because the hotfix targeted master)

### Workflow

### Merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Pre-release
 ### Fixed bugs
 - NO TICKET - add document-link back in as rich-text option
-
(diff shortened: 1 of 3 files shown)


webbyfox (approved):
LGTM

PR #974, opened 2021-02-24
Added AWS credential hook for pre-commit

webbyfox:


stevejalim (approved):
LGTM 🚢 🚢 🚢 🚢 🚢

depsiatwal (commented):
we could include this as part of test_requirements to force installation just in tests and not prod/staging/dev deployments

stevejalim:
@depsiatwal 
> we could include this as part of test_requirements to force installation just in tests and not prod/staging/dev deployments

the trick would be also running `pre-commmit install --install-hooks` too. We could probably fold that into the makefile command

webbyfox:
Yeah, I will merge this and add another PR for the makefile.

depsiatwal:
> @depsiatwal
> 
> > we could include this as part of test_requirements to force installation just in tests and not prod/staging/dev deployments
> 
> the trick would be also running `pre-commmit install --install-hooks` too. We could probably fold that into the makefile command

ahhh yes good idea

webbyfox:
> pre-commmit install --install-hooks

Adding this... DONT MERGE

stevejalim (approved):
🌟

PR #976, opened 2021-02-25
Added args for missing credential for AWS hook

webbyfox:


stevejalim (approved):
thanks for adding this. it was what i needed to make the check work when i didn't have any AWS creds file configured at all.

We should also bear in mind that there is a config flag for setting where to look for AWS creds, too, and *if* we end up using them locally from a non-standard place, the default check path might need updating

PR #979, opened 2021-02-26
Case study quote surfacing via carousel on mobile devices

rafa-garcia:
This changeset enables case study quote surfacing  on mobile devices via a sliding carousel.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1574

### Reviewing help

### Housekeeping

### Merging

image


core/sass/learn/_base.scss
```diff
@@ -890,6 +890,48 @@ $desktop: 1024px;
       height: auto;
     }
 
+    .slick-dots {
```
bobmeredith, line 893: I think you already have this styling - in product-finder.scss. Doesn't look like there is a class selector around it, so this may just be redundant.

core/templates/core/case_study_block.html
```diff
@@ -2,54 +2,31 @@
...
-        </div>
-    </div>
-</div>
+<div id="case_study"></div>
 
 {% block body_js %}
-    {% comment %} {{ block.super }} {% endcomment %}
     <script>
         let caseStudy = document.querySelector('#case_study')
+        let blocks = []
```
bobmeredith, line 10: I think these three 'let's should be const.

core/templates/core/case_study_block.html
```diff
@@ -2,54 +2,31 @@
...
-                company: caseStudy.querySelector('.case-study__company').innerText,
-                body: caseStudy.querySelector(".case-study__mobile-content--text").innerHTML,
-                carouselItems: caseStudy.querySelector(".case-study__mobile-content--media").innerHTML
-
+                heading: '{{ case_study.lead_title }}',
+                company: '{{ case_study.summary_context }}',
+                blocks: blocks,
             }
-        })
+        }, buildBlocks())
```
bobmeredith, line 29: Not quite following the logic here.  Perhaps I'm missing something but suggest you lose the global 'blocks' variable, make it local within 'buildBlocks', and return it. Also pass the list of blocks in.  Then, in the 'content' section above, you can go
`blocks: buildBlocks(caseStudy.body)`

bobmeredith (commented):
Nicely done.
Just a couple of comments

core/sass/learn/_base.scss
rafa-garcia: Thorough spotting here 👍 
However, the dots in `product-finder.scss` have a different location (top vs bottom). Also, the class `.slick-dots` is added dynamically by the `Slick` react component.

bobmeredith: OK, so how about overriding conditionally.
The beauty of .css is that the most specific selector wins, so adding the following should work:
```
.quote {
	.slick-dots {
	  bottom: -20px;
	  top: unset;
	  & li  {
	    &:before {
          background-color: transparent;
        }
	  }
	}
	.slick-list {

	  & .slick-track {
	    & .slick-slide {
	      margin-bottom: 45px;
	    }
	  }
	}
}
```
This could go in the product-finder scss .. or probably nicer to have the slick-slide in its own file perhaps.

bobmeredith:
```
.quote {
	.slick-dots {
	  bottom: -20px;
	  top: unset;
	  li  {
	    &:before {
          background-color: transparent;
        }
	  }
	}
	.slick-list .slick-track .slick-slide {
	 	margin-bottom: 45px;
	 }
}
is the same but a bit more compact actually
```

PR #992, opened 2021-03-01
GP2-1686: Porting two of the Domestic contact pages

stevejalim:
This changeset ports over the two 'basic' contact forms from BAU into Magna:
* `/contact/domestic/` - a Django view which sends to zendesk
* `/contact/domestic/enquiries/` - a Django view which sends an email via Gov UK Notify

Implementing/porting both of these meant porting a bunch of mixins and base classes, along with their tests. There are a lot, but I would really value some proper eyes on them. In terms of the rendered HTML, the pages are / should be identical to those on BAU (you can pull the code locally, migrate and run it to be sure)

Both of the views above lead to `/contact/domestic/success/`, which is a Django view that in BAU draws its content from entire Page models in domestic-cms. In Magna/great-cms, however, I've gone with a similar approach that uses Snippets instead
* Why? So that we maintain the rigour that _Pages_ in Wagtail are actual viewable standalone _pages_, so we have a page _tree_ that represents the browsable site. Because these contact pages are driven by Django views, not Wagtail, I didn't want to have to create a workflow that relied on Wagtail Page -> Django view -> Wagtail Page for instance, because this would have created a dependency on making sure certain full pages with certain slugs exist. 
* We can't get away entirely from a link between CMS content and hard-coded values in Django, I've taken a nice idea from BAU:
  * the content for this Contact Success page (and others in the future) is held in `contact.models.ContactSuccessSnippet`, where the slug/unique reference to the snippet can only be chosen from a fixed list of options:
  
Screenshot 2021-03-02 at 13 04 29

and these options are slotted into the workflow as additional url kwargs that includes the class of Snippet model to query via the ORM:
```
    path(
        'domestic/success/',
        skip_ga360(DomesticSuccessView.as_view()),
        {
            'slug': snippet_slugs.HELP_FORM_SUCCESS,
            'snippet_import_path': 'contact.models.ContactSuccessSnippet',  # see core.mixins.GetSnippetContentMixin
        },
        name='contact-us-domestic-success',
    ),
```
This means we now have a reusable pattern for CMS-held content in Magna that doesn't need us to use full Wagtail Pages for content that is rendered by a Django view.

These snippets will be able to be managed via a clear new section in the CMS (which we get for free via `wagtail-modeladmin`)

Screenshot 2021-03-02 at 12 49 44
 

Note that the `contact` app is temporarily behind a simple feature flag that means it will only appear in local dev and deployed Dev, but not Staging or Beta

### Workflow

### Reviewing help

  * pull the branch.
  * add a `ContactSucessSnippet` for 'Contact domestic form success' to your local Wagtail - see the screenshot above
  * add values for `RECAPTCHA_PRIVATE_KEY` and `RECAPTCHA_PUBLIC_KEY` to your `secrets-do-not-commit` -- if you need these, DM me and I'll share access to them with you
  * add values for 
  * you should need to set no other env vars for `great-cms` but you may do for directory-forms-api to talk to Zendesk and Notify - if you do, let me know
  * you'll need to be running directory-forms-api locally too
  * at `/contact/domestic/` when you submit a message it will appear in the admin of directory-forms-api and will also go to the dev/staging zendesk - DM me for where to find this if you don't know already
  * at `/contact/domestic/enquiries/` submissions go into directory-forms-api and also get the CI Mailhog / mail trap. You'll get a confirmation email from Notify, regardless, so please use your real work email address as the person sending to see this.
  * confirm that `/contact/domestic/success/` contains the content you added to the CMS earlier.

### Housekeeping

- [x] Added all new environment variables to Vault - WIP - will tick when done

### Merging


contact/constants.py
```diff
@@ -0,0 +1,14 @@
+LIMITED = 'LIMITED'
```
depsiatwal, line 1: We have directory constants to old these

depsiatwal (commented):
Generally LGTM at high level I notice some tests may need to be added to increase coverage

contact/models.py
```diff
@@ -0,0 +1,157 @@
...
+from django.forms import Select
+from wagtail.admin.edit_handlers import FieldPanel, MultiFieldPanel
+from wagtail.snippets.models import register_snippet
+
+from core.cms_panels import SearchEngineOptimisationPanel
+from . import snippet_slugs
+
+
+class NonPageContentSnippetBase(models.Model):
+    # Ensures all ContentSnippets have the minimum reqired fields, such as `slug`
```
webbyfox, line 11: typo

webbyfox (commented):
LGTM

webbyfox (approved):

contact/constants.py
stevejalim: Ah good call - will clean up in a follow-up piece around this. I guess directory-constants is a repo that will remain around because it's used by directory-api as well? if it's only used by directory-cms and great-domestic-ui (and international) then we can eventually retire it

PR #1008, opened 2021-03-05
Gp2 1548 service removed page

Rambowalsh:
- [x ] Ticket exists in Jira https://uktrade.atlassian.net/browse/Gp2-1548
- [x ] Jira ticket has the correct status.
- [x ] [Changelog](CHANGELOG.md) entry added.

### Housekeeping

### Merging


core/urls.py
```diff
@@ -20,6 +20,9 @@ def anonymous_user_required(function):
 
 
 urlpatterns = [
+    re_path(r'^triage/(?P<step>.+)/$', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='triage-wizard'),
+    re_path(r'^triage/$', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='triage-start'),
+    re_path(r'^custom/$', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='custom-page'),
```
stevejalim, line 25: It would be nice to change these from be BAU-style `re_path()` to be just `path() like the rest of the file (the syntax changed a while back, but after BAU was created https://docs.djangoproject.com/en/2.0/releases/2.0/#whats-new-2-0)

I think these should work, but give it a test drive locally to be sure:

    path('triage/<slug:step>/', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='triage-wizard'),
    path('triage/', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='triage-start'),
    path('custom/', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='custom-page'),

core/urls.py
```diff
@@ -51,6 +54,7 @@ def anonymous_user_required(function):
         ),
         name='signup-wizard-export-plan',
     ),
+    path('service-unavailable/', skip_ga360(views.ServiceNoLongerAvailableView.as_view()), name='service-unavailable'),
```
stevejalim, line 57: This is useful for testing, but I can't remember if the story asked for a specific URL for the Service No Longer Available view. If it didn't, we should probably delete this entry

core/views.py
```diff
@@ -277,3 +277,15 @@ def form_valid(self, form):
 
 class ContactUsHelpSuccessView(TemplateView):
     template_name = 'core/contact-us-help-form-success.html'
+
+
+class ServiceNoLongerAvailableView(TemplateView):
+    slug = 'advice'
```
stevejalim, line 283: This line (a hangover from BAU) isn't needed - it repeats what we set as `advice_page_slug` below

```suggestion
```

tests/unit/core/test_views.py
```diff
@@ -88,6 +88,17 @@ def contact_form_data(captcha_stub):
...
 
 
+@pytest.mark.django_db
+def test_service_removed_page(
+    client,
+    domestic_homepage,
+    domestic_site,
+):
+    # get a path we know will trigger the service-removed view
+    response = client.get('/triage/foo/')
```
stevejalim, line 98: Thanks for this. Would be good to test the other two URLs that are wired in - so just `/triage/` and also `/custom/` too, please

stevejalim (changes requested):
Looking good - just a couple of small changes (nothing mega) and we can get this merged

core/urls.py
Rambowalsh: Amended and tested local

core/views.py
Rambowalsh: Amended

tests/unit/core/test_views.py
Rambowalsh: amended

core/urls.py
Rambowalsh: Amended

stevejalim (approved):
Just realised it'd be ideal to expand the test coverage to also set up some child pages of a TopicLandingPage that has the slug `advice` and confirm they are shown in the resulting rendered page.

If you want to add that so we get extra protection in the future, that's great, but approving this PR as-is because it works. I'd prefer us to have the extra test, but it's not a blocker

Rambowalsh:
All approved service removed page

PR #1012, opened 2021-03-05
GP2-1778: Support post auth redirect for Wagtail and Django views

stevejalim:
This changeset updates our sign-in and sign-up flows so that they can send a user on to their originally requested destination (if relevant) after authentication. 

Some notes:

* Previously, everyone was automatically sent to /dashboard/ after authentication, but now we respect any `?next=` querystring for both Wagtail views and all Django Views.
* Also, if someone lands on `/signup/` with a `?next=` querystring param then goes to `/login/`, the `next` param is carried across too. Previously, this only happened for the login URL.
* The default page for requiring auth is now the SIGN UP page, not login, because data shows that the majority of our users don't have an account yet/currently. In the future we may swing this back to being login as the default. 
    * The 'Sign In' link in the header for the Great Domestic Homepage, however, still sends users to login, to match the intent expressed in the words 'Sign in'
 * A template tag has been added that does still sends people to `/dashboard/` if they don't have a deliberate onward destination in `?next`. It also does some light sense-checking on the value of `next` so we can avoid sending people to /login/ after login, for instance, to onwards to a different URL etc

### Workflow

### Reviewing help

* If you don't have  a GreatDomesticHomepage set up locally, add one at /domestic/
* in the first section of CTAs, add links to the following pages, both of which require authentication
    *  http://greatcms.trade.great:8020/learn/categories/getting-started/lesson-basics/lesson-same-files-used-twice/
    * http://greatcms.trade.great:8020/export-plan/section/marketing-approach/    
* Publish the page and view it in an incognito window where you are not signed in
* For each of those CTA links:
   * Click it and follow the Sign Up flow, including the email verification code workflow. You will end up on the desired target page in the end
   * Log out
 * Repeat the above but instead of signing in, just log back in with the credentials you created. Again, you'll end on on the desired target page in the end  

### Merging


core/templates/components/header_footer/sso_login.html
```diff
@@ -1,6 +1,8 @@
-{% if sso_is_logged_in %}
+{% load wagtailcore_tags %}
+{% if not sso_is_logged_in %}
```
stevejalim, line 2: @lewis-coulson-dit  Does this `not` look right to you? I think it disappeared from develop after I had started this branch and so now i am reinstating it

exportplan/urls.py
```diff
@@ -4,83 +4,91 @@
...
+        'service-page/',
+        login_required(views.ExportPlanServicePage.as_view(), login_url=SIGNUP_URL),
+        name='service-page',
     ),
-    path('pdf-download/', login_required(views.PDFDownload.as_view(), login_url=LOGIN_URL), name='pdf-download'),
+    path('pdf-download/', login_required(views.PDFDownload.as_view(), login_url=SIGNUP_URL), name='pdf-download'),
     path(
         'api/recommended-countries/',
-        login_required(skip_ga360(api.ExportPlanRecommendedCountriesDataView.as_view()), login_url=LOGIN_URL),
+        login_required(skip_ga360(api.ExportPlanRecommendedCountriesDataView.as_view()), login_url=SIGNUP_URL),
```
bobmeredith, line 91: This is a bit repetitive. Could we use LOGIN_URL in settings.py to do the same thing?  ATM, that is set as 'SSO_PROXY_LOGIN_URL' but I guess that's not a sensible setting for Magna.

stevejalim: in settings.py `LOGIN_URL` is kinda hijacked to specifically be the Wagtail login, either via username and password (local dev) or SSO, so I left LOGIN_URL and SIGNUP_URL as a non-settings variables. Happy for me to leave as-is? Don't really want to muddy the waters of Admin SSO login @bobmeredith  😄

bobmeredith (approved):
LGTM

PR #1017, opened 2021-03-08
GP2-1569 Compare markets age group tab

bobmeredith:
Adds a 'compare age groups' tab to the data table in 'Where to export'.
This table allows dynamic filtering. Changing of filters does not require a round-trip to the B/E as all the data are retrieved on first load and table updated from calculations.
These data are retrieved using the generic 'country data' api request. Since the data loaded for population-by-age-group cover a wide range of years - going forward to 2100, it was necessary to add filtering to the api mechanism so that only the correct year is retrieved. This filtering is applied to all models which gives up a simple implementation for showing trends at a later stage. Any arbitrary django filter can be applied in fact, and on any model. There are no security issues with this as a) only models from the dataservices application can be retrieved, and b) Only ORM filters can be applied, thus preventing access to raw sql.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1569

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
Testing: Go to the 'Where to export' page Observe 'Target age groups' tab and select.
The data should load showing all age groups. On selecting any filter checkbox, the table should respond instantly, showing only the data for the selected age group.
Not Italy has no population age range data currently due to a mis-match between the numeric iso code in our reference data, and the er-hem correct one.

bobmeredith:
image

package.json
```diff
@@ -98,7 +98,7 @@
     "@babel/polyfill": "^7.12.1",
     "@fortawesome/fontawesome-free": "^5.15.1",
     "@storybook/react": "^6.0.16",
-    "great-styles": "github:uktrade/great-styles",
+    "great-styles": "*",
```
dahfool, line 101: @bobmeredith This will break CI.

dahfool (approved):

PR #1019, opened 2021-03-09
Gp2 1576 casestudy csv

webbyfox:
This changeset adds a download button in admin UI to export all case studies attributes into CSV.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/1576

On the case study admin index page, we have download csv button 

image


CHANGELOG.md
```diff
@@ -32,15 +36,6 @@
 - GP2-1735 - update lessons
 - GP2-1632-generic-ep-pages
 - GP2-1782 - ep section progress
-
```
stevejalim, line 35: Thanks for cleaning up this 💅

cms_extras/modeladmin.py
```diff
@@ -1,13 +1,16 @@
...
 from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
+from wagtail.contrib.modeladmin.views import IndexView
 
 from core.models import CaseStudy
 
 
 class CaseStudyAdminButtonHelper(ButtonHelper):
-
-    view_button_classnames = ['button-small', 'icon', 'icon-doc']
+    view_button_classnames = ['button-small', 'icon', 'icon-doc', 'btn-group']
```
stevejalim, line 13: Just curious - what's this `btn-group` for?

cms_extras/modeladmin.py
```diff
@@ -26,12 +29,97 @@ def get_buttons_for_obj(self, obj, exclude=None, classnames_add=None, classnames
...
+                for field in self.list_export
+                if field not in self.columns_to_convert
+            }
+        )
+
+        for item in queryset:
+            yield self.write_csv_row(writer, self.to_row_dict(item))
+
+    def create_row_by_tag(self, row_dict, processed_row):
+        processed_row_list = []
```
stevejalim, line 61: I'd add a comment here explaining that this expands the stub of a single processed row into its various permutations, just to help our future selves

cms_extras/modeladmin.py
```diff
@@ -26,12 +29,97 @@ def get_buttons_for_obj(self, obj, exclude=None, classnames_add=None, classnames
...
+        for f in self.columns_to_convert:
+            # No tagged value then just append row for write
+            attr_values = [] if row_dict.get(f) == '-' else row_dict.get(f)
```
stevejalim, line 64: tiny thing - is it worth moving this to be the 'else' of an 'if / elif / else  ' below to avoid wasted computation if `f` is `'get_related_pages'` or `'modified'`?

cms_extras/modeladmin.py
```diff
@@ -26,12 +29,97 @@ def get_buttons_for_obj(self, obj, exclude=None, classnames_add=None, classnames
...
+
+    def write_csv_row(self, writer, row_dict):
+        processed_row = {}
+        common_fields = {k: v for k, v in row_dict.items() if k in ['title', 'summary_context', 'lead_title']}
+        for field, value in common_fields.items():
+            preprocess_function = self.get_preprocess_function(field, value, self.FORMAT_CSV)
+            processed_value = preprocess_function(value) if preprocess_function else value
+            processed_row[field] = processed_value
+        return self.write_multiple_rows(writer, self.create_row_by_tag(row_dict, processed_row))
+
```
stevejalim, line 95: Not a change request but this whole group of methods feels like it'd benefit from type hinting on the signatures - what do you think?

cms_extras/modeladmin.py
```diff
@@ -81,5 +169,11 @@ def get_related_pages(self, obj):
 
     get_related_pages.short_description = 'Associated pages'
 
+    def association(self, obj):
+        pass
+
+    def attribute(self, obj):
+        pass
+
```
stevejalim, line 177: Why are these here if they don't (I think?) render anything in the modeladmin?

config/settings.py
```diff
@@ -9,6 +9,7 @@
 from .utils import get_wagtail_transfer_configuration
 
 ROOT_DIR = environ.Path(__file__) - 2
+APPS_DIR = ROOT_DIR.path('core')
```
stevejalim, line 12: I think this could be better named as `CORE_APP_DIR` because `APPS_DIR` sounds like it's all the apps not just `core`

config/settings.py
```diff
@@ -95,9 +96,15 @@
...
-        'APP_DIRS': True,
+        'DIRS': [
+            APPS_DIR.path('templates'),
+        ],
         'OPTIONS': {
             'debug': True,
+            'loaders': [
+                'django.template.loaders.filesystem.Loader',
+                'django.template.loaders.app_directories.Loader',
+            ],
```
stevejalim, line 107: Am surprised we needed to add these - is this specifically to get the CSV export to work?

core/templates/modeladmin/index.html
```diff
@@ -0,0 +1,101 @@
...
+                    <div class="right">
+
+                        {% if user_can_create %}
+                            <div class="actionbutton">
+                                {% include 'modeladmin/includes/button.html' with button=view.button_helper.add_button %}
+
+
+                        {% if view.list_export %}
+                            <div class="dropdown dropdown-button match-width col">
+                                <a href="?export=csv&{{ request.GET.urlencode }}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Download CSV' %}</a>
```
stevejalim, line 36: Is this the only line that's customised? If so, It'd be better to only override `modeladmin/index.html` for `{% block header_extra %}` and reduce the duplication in this template. 

If there are more changes in this template, would be cool to make them easy to find via comments, in case someone overwrites them if, say, the `wagtail.contrib.modeladmin` is refactored

stevejalim (commented):
Looking pretty great - nice one! ⭐ 

Added a bunch of small questions and two comments that might need attention (the ones about `APPS_DIR` and `modeladmin/index.html`

Let me know what you think and am happy to approve after that.

cms_extras/modeladmin.py
webbyfox: Added comment in code for this

core/templates/modeladmin/index.html
webbyfox: Thanks for picking this. Good catch

stevejalim (approved):
LGTM - thanks!

I think we could put _more_ type hinting on some signatures, but it's not a blocker at all 🚢

PR #1020, opened 2021-03-09
Fix seo title not appearing in meta tags tkt GP2-1811

Rambowalsh:
### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/TGP2-1811

### Merging


domestic/templates/domestic/base.html
```diff
@@ -12,6 +12,9 @@
         {% block sharing_description %}
             <meta name="description" content="{% block meta_description %}{% firstof page.search_description page.teaser page.featured_description page.seo_title page.title content_snippet.search_description %}{% endblock %}">
         {% endblock %}
+        {% block sharing_title %}
+            <meta name="title" content="{% block meta_title %}{% firstof page.seo_title page.teaser page.featured_description page.seo_title page.title content_snippet.search_description %}{% endblock %}">
```
bobmeredith, line 16: You seem to have ```page.seo_title``` in here twice.

domestic/templates/domestic/base.html
```diff
@@ -12,6 +12,9 @@
...
+        {% endblock %}
```
stevejalim, line 17: I don't think we would want the other values to be used as the fallback for the `meta name='title'` tag - `page.teaser page.featured_description page.title content_snippet.search_description` etc

Looking at the ticket, I think it's worth clarifying if the content team want the SEO title to appear only in the `meta`, or to be used as the actual page `<title>` too.

stevejalim (changes requested):
I think it's worth a tweak and also a double check what the content team need/are specifically asking for - see my comment. Cheers!

domestic/templates/domestic/base.html
stevejalim: Also, I'm not 100% sure if `meta name='title'` is still current/viable. I find it in some older web discussions, but not in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name for instance

Assuming it's no longer viable, I think that changes scope of this more to:
* set a sensible default (using the SEO title field from the CMS) for the base template's <title> tag
* and also to tweak the `country_guide.html` template so that it uses it this default instead of what's there now.

if so, on this the base template, I'd tweak:

* `{% firstof page.seo_title page.title %}` as the  _default_ value onside the head_title block (see the <title> tags a few lines further down 

then in the template for the country_guide.html template:

*  remove `{% block head_title %} {% endblock %}`  from `country_guide.html` altogether so we just get the default behaviour

* tell the content team that whatever they put in SEO title or page title will now appear in the <title> tag

stevejalim: Ah , `meta name="title"` might be fine, yeah - it's globally allowed attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) but I would still check in with Content to see if they want control over the actual <title> tag and not just the `meta`

domestic/templates/domestic/base.html
Rambowalsh: Resolved

domestic/templates/domestic/base.html
Rambowalsh: Amended. I out the change in to use the SEO title over the page title as well If you can ok that

domestic/templates/domestic/base.html
```diff
@@ -12,6 +12,9 @@
         {% block sharing_description %}
             <meta name="description" content="{% block meta_description %}{% firstof page.search_description page.teaser page.featured_description page.seo_title page.title content_snippet.search_description %}{% endblock %}">
         {% endblock %}
+        {% block sharing_title %}
+            <meta name="title" content="{% block meta_title %}{% firstof page.seo_title%}{% endblock %}">
```
stevejalim, line 16: `firstof` usually takes multiple options for what to pick, so if we've just got one, there's a risk we'll end up with a `meta` tag for `title` that has no content.

I'd either put the whole tag in a in {% if page.seo_title %} clause, or add page.title to the list of options for the `firstof` tag -- and I'd go with whichever matches the behaviour that decides what goes in the main <title> tag, i think (but see if Team Content have a view on this)

domestic/templates/domestic/base.html
```diff
@@ -23,7 +26,13 @@
           <meta id="privacyCookieDomain" value=".great.gov.uk" />
         {% endblock %}
         <title>
-            {% block head_title %}{% endblock %}
+            {% block head_title %}
+                {% if page.seo_title %}
+                    {{page.seo_title}}
+                {% else %}
+                    {{page.title}}
```
stevejalim, line 33: I can see that other pages - eg article_page.html have this as their `head_title` content:

{% block head_title %}{{ page.<name of title field here> }} - great.gov.uk{% endblock %}

so I think it's worth adding that - `- great.gov.uk` suffix to things:
```suggestion
                    {{page.title}} - great.gov.uk
```

stevejalim (changes requested):
heya - this is a bit of a rabbit warren of a ticket, I'm afraid.

Got some suggested changes + there are a couple more templates in the codebase that have hard-coded `head_title` blocks that might be better as customisable. Check with Team Content / Adam to see if they want them editable as SEO_title etc:

* landing_page.html (used for "Advice" and "Success Stories" I think)
* performance_dashboard.html

Cheers!

domestic/templates/domestic/base.html
```diff
@@ -12,6 +12,9 @@
         {% block sharing_description %}
             <meta name="description" content="{% block meta_description %}{% firstof page.search_description page.teaser page.featured_description page.seo_title page.title content_snippet.search_description %}{% endblock %}">
         {% endblock %}
+        {% block sharing_title %}
+            <meta name="title" content="{% block meta_title %}{% firstof page.seo_title page.title%}- great.gov.uk{% endblock %}">
```
stevejalim, line 16: Small thing: def need a space before `- great.gov.uk` 

```suggestion
            <meta name="title" content="{% block meta_title %}{% firstof page.seo_title page.title %} - great.gov.uk{% endblock %}">
```

domestic/templates/domestic/guidance_page.html
```diff
@@ -3,7 +3,7 @@
 {% load static from staticfiles %}
 {% load breadcrumbs from component_tags %}
 
-{% block head_title %}{{ page.title }}{% endblock %}
+{% block head_title %}{% firstof page.seo_title page.title%} - great.gov.uk{% endblock %}
```
stevejalim, line 6: I don't _think_ it'll blow up without a space before the `%` but can't be 100% sure off the top of my head. Either way, worth adding one in for consistency with other tags

```suggestion
{% block head_title %}{% firstof page.seo_title page.title %} - great.gov.uk{% endblock %}
```

domestic/templates/domestic/landing_page.html
```diff
@@ -15,7 +15,7 @@
   <script src="{% static 'js/home.js' %}"></script>
 {% endblock %}
 
-{% block head_title %}Welcome to great.gov.uk - export guidance and services{% endblock %}
+{% block head_title %}{% firstof page.seo_title page.title%}- great.gov.uk{% endblock %}
```
stevejalim, line 18: Again, my paranoia about trailing spaces before `%}` 😄 

```suggestion
{% block head_title %}{% firstof page.seo_title page.title %}- great.gov.uk{% endblock %}
```

stevejalim (changes requested):
Looking good - just a couple of tiny requests involving whitespace, only one of which is def needed, though.

domestic/templates/domestic/guidance_page.html
Rambowalsh: Ok done

domestic/templates/domestic/base.html
Rambowalsh: Done

domestic/templates/domestic/base.html
Rambowalsh: Done

domestic/templates/domestic/base.html
Rambowalsh: Done

domestic/templates/domestic/base.html
```diff
@@ -12,6 +12,9 @@
         {% block sharing_description %}
             <meta name="description" content="{% block meta_description %}{% firstof page.search_description page.teaser page.featured_description page.seo_title page.title content_snippet.search_description %}{% endblock %}">
         {% endblock %}
+        {% block sharing_title %}
+            <meta name="title" content="{% block meta_title %}{% firstof page.seo_title page.title%} - great.gov.uk{% endblock %}">
```
stevejalim, line 16:
```suggestion
            <meta name="title" content="{% block meta_title %}{% firstof page.seo_title page.title %} - great.gov.uk{% endblock %}">
```

stevejalim (approved):
LGTM - cheers!

PR #1032, opened 2021-03-10
GP2-1842: Fix absent href value in service-removed page

stevejalim:
Quick fix for dev for the Service Removed page - the template still contained reference to variables that only exist on BAU, not in Magna.

Ideally this would have been caught in a test, but I remember at the time saying it likely didn't need one, so I've added it now, as penance

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1842

### Merging


domestic/templates/domestic/service_no_longer_available.html
```diff
@@ -14,7 +14,7 @@ <h1 class="heading-xlarge">This service has been removed</h1>
         <p>Read about how to:</p>
         <ul class="list list-bullet">
           {% for listing in listing_page.child_pages.live %}
-            <li><a href="{{ listing.full_path }}" class="link">{{ listing.title }}</a></li>
+            <li><a href="{{ listing.url }}" class="link">{{ listing.title }}</a></li>
```
stevejalim, line 17: `full_path` is how the data was (re)structured by the time it reached `great-domestic-ui` but in `great-cms` `listing` is just a Page instance, so we use its `.url` property instead

PR #1040, opened 2021-03-12
GP2-1775: advice page banner

lewis-coulson-dit:
Tweaks previous implementation to work with all "Advice" parent pages 

No changelog needed as GP2-1775 already added

### Merging


domestic/templatetags/component_tags.py
```diff
@@ -225,5 +227,16 @@ def success_box(**kwargs):
...
-    return child_page.get_parent().slug == slug_
+def is_descendant_of_parent_with_slug(child_page, parent_type_classname, slug_):
+
+    parent_class = {
+        'TopicLandingPage': TopicLandingPage,
+        'ArticleListingPage': ArticleListingPage,
+    }.get(parent_type_classname)
+
+    if parent_class:
+        for page_ in child_page.get_ancestors().specific():
```
webbyfox, line 238: Do we need to check page is live or not?

lewis-coulson-dit: Guessing not since this will only be used on live pages anyway. @stevejalim is this ok?

domestic/templatetags/component_tags.py
```diff
@@ -225,5 +227,16 @@ def success_box(**kwargs):
 
 
 @register.simple_tag
-def is_child_of_parent_with_slug(child_page, slug_):
-    return child_page.get_parent().slug == slug_
+def is_descendant_of_parent_with_slug(child_page, parent_type_classname, slug_):
```
webbyfox, line 230: This tag name is very generic where we check only for only two parent page types. Not a blocker though

webbyfox (approved):
Minor comment and question

domestic/templatetags/component_tags.py
stevejalim: The idea is we might/will need to expand its scope in future. Alternatively, we could refactor it to just compare class names as strings, so then it'll work for everything...

domestic/templatetags/component_tags.py
stevejalim: I wondered about live() or not but we might end up making a parent page draft to be able to use the parent as a purely structural part of a slug with no content of its own, so .live() would break this tag in that situation.

So, yeah, i think it's OK but thanks for asking @webbyfox  🙌

PR #1041, opened 2021-03-12
GP2-1724 Speed up population tab

bobmeredith:
This PR results from a spike looking at loading performance on the 'Where to export' tab.
the population tab used a tab-specific endpoint that was implemented in a very slow manner in directory api. Simply re-configuring the tab to use the newer bulk-request endpoint, accessing db models in directory-api, made the tab performance similar to other tabs.
There was one missing model in directory-api (urban/rural population) but with this added, the old population endpoint could be abandoned.
Also, there is some work to remove redundant endpoints and the unnecessary (and completely unused) loading of local .csv files, along with removing said files from fixtures.  These data are available from directory-api now.
A couple of minor bugs:
- Keep selected filters on age-group tab when switching between tabs.
- Use the maximum EODB ranking as the 'out of' rather than the count of DB rows as the latter is wrong.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1724

### Reviewing help

No screenshot as the UI is unchanged.  Should be faster though.

### Housekeeping

### Merging


bobmeredith:
Testing.
Check for regression (particularly on export plan data points) and tabs in WTE. The population tab should load much faster than before. Also, go to age-group tab, select a filter - change tab then back again. Check that the correct filters are ticked and that the data are being filtered correctly.

webbyfox (approved):
Welldone @bobmeredith

PR #1052, opened 2021-03-16
GP2-1563 Input clear button on mobile

bobmeredith:
There are still issues using the clear button on search field in product finder on mobile. This PR attempts a fix based only on video evidence.

Code change explanation:  It looks like the click event fires but the input is not re-focussed and the content doesn't clear which leads me to suspect that the 'isfocussed' flag is not at the right state - perhaps clearing if the focus-out event occurs on the input without active element being set.
Anyway, the flag check is not really needed as the clear button is only populated if the state is correct.

### Workflow

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
Testing.  On mobile - go to product finder. Edit the product name, tap the 'clear' cross.
the input content should clear, the cross replaced by pen icon, and the input should remain focussed.
Check that the tap area is sufficient.

lewis-coulson-dit (approved):

PR #1068, opened 2021-03-18
Remove banner logged out condiional and fix incorrect link

Zak-JS:
### Merging

- [x] This PR can be merged by reviewers. (If unticked, please leave for the author to merge)

PR #1069, opened 2021-03-18
Release Feature/banner auth link fix

Zak-JS:
Banner auth requirement removal

### DIFF
--- a/domestic/templates/domestic/article_listing_page.html
+++ b/domestic/templates/domestic/article_listing_page.html
@@ -18,7 +18,7 @@
     {% endblock %}
   </div>
 
-  {% if 'advice' in page.get_parent|lower and not request.user.is_authenticated  %}
+  {% if 'advice' in page.get_parent|lower %}
     <div class="container">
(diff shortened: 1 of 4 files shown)


domestic/templates/domestic/article_listing_page.html
```diff
@@ -18,7 +18,7 @@
     {% endblock %}
   </div>
 
-  {% if 'advice' in page.get_parent|lower and not request.user.is_authenticated  %}
```
stevejalim, line 21: How come this line isn't in the #1068 ? Was it already on `develop`?

stevejalim (commented):
Quick question, then can approve once #1068 is tested on dev

domestic/templates/domestic/article_listing_page.html
stevejalim: Resolved: same change is now heading to develop via #1068

stevejalim (approved):
LGTM - BUT ONLY safe to merge after DM/PO signoff on dev first

PR #1070, opened 2021-03-18
GP2-1900: Port URL redirects with tests from BAU

Rambowalsh:
### Workflow

- [x ] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1900
- [x ] Jira ticket has the correct status.
- [x ] [Changelog](CHANGELOG.md) entry added.

### Merging


Rambowalsh:
first test fails due to lots of tests having to be commented out as not all requirements for trhem to run are in magna yet

config/url_redirects.py
```diff
@@ -0,0 +1,807 @@
...
+            url='https://www.events.great.gov.uk/ehome/trade-events-calendar/all-events?keyword=australia'
+        ),
+        name='australia-event-calendar',
+    ),
+    url(
+        r'^trskorea/$(?i)',
+        QuerystringRedirectView.as_view(url='https://eu.eventscloud.com/korea-uk-tech-rocketship-awards-kr'),
+    ),
+    url(
+        r'^TRSkorea/$',
```
webbyfox, line 74: Do we need this twice? It's just a case that can be handle on regex. I think #L70 you've case insensitive flag, is it not working?

config/url_redirects.py
```diff
@@ -0,0 +1,807 @@
...
+            url='/international/',
+            language=redirect[1],
+        ),
+    )
+    for redirect in (INTERNATIONAL_LANGUAGE_REDIRECTS_MAPPING + INTERNATIONAL_COUNTRY_REDIRECTS_MAPPING)
+]
+
+# TOS and privacy-and-cookies are no longer translated, instead we redirect to
+# the ENG version
+TOS_AND_PRIVACY_REDIRECT_LANGUAGES = ('zh', 'ja', r'es', 'pt', 'ar', 'de')
```
webbyfox, line 396: `r'es'`?

stevejalim: Not sure if you've spotted a bug but that's the same as it is in BAU/V1

config/url_redirects.py
stevejalim: We're porting directly from V1, so it might be that this IS needed in a subtle way. Will try locally shortly

config/url_redirects.py
webbyfox: Just wondering why `es` been treated as a raw string and not other - I think normal string will be OK as r string use for translating various escape codes which is not present here

config/url_redirects.py
stevejalim: L70 seems to do the job so removing L73:76

config/url_redirects.py
stevejalim: oh yeah trying wihout `r`

webbyfox (approved):
LGTM Thanks @stevejalim @Rambowalsh

PR #1071, opened 2021-03-18
Bumping wagtail-transfer package

webbyfox:

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -18,8 +18,6 @@ appdirs==1.4.4
     # via
     #   black
     #   virtualenv
-appnope==0.1.2
-    # via ipython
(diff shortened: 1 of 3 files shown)


requirements_test.txt
```diff
@@ -18,8 +18,6 @@ appdirs==1.4.4
     # via
     #   black
     #   virtualenv
-appnope==0.1.2
-    # via ipython
```
stevejalim, line 22: I wonder why this disappeared...

stevejalim (approved):

requirements_test.txt
webbyfox: dont know TBH, maybe `ipython` has removed it

PR #1073, opened 2021-03-19
New modal for segmentation question

rafa-garcia:
CONTEXT: This PR adds a modal for allocating users into discrete segments so that we can better support their needs based on their response. The changeset includes a refactor to the current Tour component along with a new React component to manage the launch of both modals.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1719

### Reviewing help

- [X] Includes screenshot

### Housekeeping

### Merging

image


react-components/src/components/PageModalManager.jsx
```diff
@@ -0,0 +1,42 @@
...
+export function PageModalManager(props) {
+  const { modals } = props
+  const [currentModal, setCurrentModal] = useState(modals[0])
+
+  const nextModal = () => {
+    const currentIndex = modals.indexOf(currentModal)
+    setCurrentModal(modals[currentIndex + 1] || modals[currentIndex])
+  }
+
+  let Component
```
bobmeredith, line 14: You can declare and assign in one `let component = xxxxxx`
Also, variables should be camelCase

react-components/src/components/Segmentation/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+import ReactModal from 'react-modal'
+import Services from '@src/Services'
+import Interaction from './Interaction'
+
+export default function Segmentation(props) {
+  const { segment, handleModalClose } = props
+  const [isOpen, setIsOpen] = useState(!!segment)
+  const [isInProgress, setIsInProgress] = useState()
+
+  const segmentQuestion = {
```
bobmeredith, line 12: It looks like this is combining the question mechanics with the data. I'm thinking this should be elsewhere, possibly even loaded through a separate request.
It's not too much of a problem as yet given the one-question requirement here. We could refactor later.

react-components/src/components/Segmentation/Interaction.jsx
```diff
@@ -0,0 +1,105 @@
+import React, { useState } from 'react'
+import PropTypes from 'prop-types'
+
+function RadioButtons(props) {
```
bobmeredith, line 4: Would it maybe possible to use the same RadioButtons component as product finder. I didn't split it from the Interaction file, could it could easily be done I guess.

bobmeredith (approved):
Nice one.
Yeah - a lot of test fallout because of the user profile thing.
I left a couple of comments, but could easily be addressed in the refactor that I know is crouching on the horizon.

PR #1074, opened 2021-03-19
Wagtail upgrade 2.11

webbyfox:
This changeset upgrade wagtail to 2.11.*

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -81,7 +81,7 @@ MIDDLEWARE = [
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'whitenoise.middleware.WhiteNoiseMiddleware',
-    'wagtail.core.middleware.SiteMiddleware',
+    'wagtail.contrib.legacy.sitemiddleware.SiteMiddleware',
     'wagtail.contrib.redirects.middleware.RedirectMiddleware',
     'core.middleware.UserSpecificRedirectMiddleware',
     'core.middleware.UserLocationStoreMiddleware',
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -6,7 +6,7 @@ from django.conf import settings
 from django.core.files import File
 from django.core.files.base import ContentFile
 from django.urls import clear_url_caches
-from wagtail.core.models import Collection
+from wagtail.core.models import Collection, Locale
 from wagtailmedia import models as wagtailmedia_models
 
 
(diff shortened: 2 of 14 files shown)


config/settings.py
```diff
@@ -81,7 +81,7 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'whitenoise.middleware.WhiteNoiseMiddleware',
-    'wagtail.core.middleware.SiteMiddleware',
+    'wagtail.contrib.legacy.sitemiddleware.SiteMiddleware',
```
stevejalim, line 84: Ideally we should (later, not today) replace this legacy middleware with the alternative approach that doesn't need middleware at all

tests/helpers.py
```diff
@@ -47,3 +47,13 @@ def make_test_video(
...
 
     return media
+
+
+class SetUpLocaleMixin:
+    """Ensures a default locale is in the database
+    pytest equivalent is the `en_locale` fixture
+    """
+
+    def _fixture_setup(self):
```
stevejalim, line 57: 💯  for spotting the appropriate method to extend @webbyfox  🌟

stevejalim (approved):
Tried it locally and LGTM

tests/unit/domestic/test_models.py
```diff
@@ -54,7 +54,7 @@
 )
 
 
-class DomesticHomePageTests(WagtailPageTests):
+class DomesticHomePageTests(SetUpLocaleMixin, WagtailPageTests):
```
depsiatwal, line 57: SetupLocal 
wondering if it's easier and cleaner to add as fixture that's always included ?

depsiatwal (commented):
generally LGTM small comment

tests/unit/domestic/test_models.py
stevejalim: @depsiatwal Thanks - I appreciate the suggestion.

We have added a pytest fixture (`en_locale`) and used it where needed in pytest tests, but setting `autouse=True` requires updating 300 tests with `@pytest.mark.django_db` even if the test itself doesn't need the en_locale fixture or DB access. So instead we've gone with targeted use of the fixture (explict > implicit and all that).

As a result, and because unittest.TestCase only likes `autouse`-d fixtures, we had to make a mixin class to do the same kind of job the fixture would otherwise have done.

Are you cool with that? Happy to discuss more

PR #1079, opened 2021-03-23
GP2-1840: signed in header

lewis-coulson-dit:
New logged in Magna header.

image

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1840

### Merging


react-components/src/components/ModalMenu.jsx
```diff
@@ -7,7 +7,7 @@ import PropTypes from 'prop-types'
 const customStyles = {
   overlay: {
     zIndex: '3',
-    background: 'transparent',
+    background: 'rgba(0, 0, 0, 0.6)',
```
bobmeredith, line 10: Seriously? When does a TRH menu ever darken the screen?

react-components/src/components/ModalMenu.jsx
```diff
@@ -20,15 +20,18 @@ export function Menu(props) {
...
+  const firstMenuItem = useRef(null)
+  const lastMenuItem = useRef(null)
+  const menuItem = useRef(null)
 
   const openModal = (evt) => {
     const position = evt.target.getClientRects()[0] || { top: 0, height: 0 }
     const bodyWidth = evt.target.closest('body').clientWidth
     customStyles.content.top = `${
-      position.top + position.height + window.scrollY
+      (position.top + position.height + window.scrollY) - 15
```
bobmeredith, line 31: Not sure why you need these constants. The menu is now placed directly under the button with its RHS aligned to the right edge.

bobmeredith (approved):
Looks OK, although it seems like a sad downgrade. What the 
Not sure why you need the positioning constants - it would be maybe nicer to move those into the .css as margins instead perhaps

react-components/src/components/ModalMenu.jsx
lewis-coulson-dit: In the lovely new design it does :)

react-components/src/components/ModalMenu.jsx
lewis-coulson-dit: Good spot, yes my approach was to change as little as possible with the existing header code since this work is temporary.

PR #1080, opened 2021-03-23
Fix progression if initial modal state is close

rafa-garcia:


react-components/src/components/Segmentation/index.jsx
```diff
@@ -96,6 +102,6 @@ export default function Segmentation(props) {
 }
 
 Segmentation.propTypes = {
-  segment: PropTypes.string.isRequired,
```
bobmeredith, line 99: I think, to make eslint happy, all non-mandatory prop-types should have a default.
so Segmentation.defaultProps = {segment:''} - maybe

bobmeredith (approved):
LGTM

PR #1091, opened 2021-03-24
Add min date available to select for both to and from, add friendly date format.

jakehopking:
CONTEXT: This changeset adds enhancements to input of type=date:

- Start date should be today's date and unable to select before this date
- End date shouldn't be before the start date 
- Date should be in friendly short format e.g.: 15 Oct 2021
- Adds helper to generate friendly short format date.
- Adds helper to generate ISO format date.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-800

### Reviewing help

- [X] Test with `npx jest -t dateFormat`

### Housekeeping

### Merging

### After

https://user-images.githubusercontent.com/3432643/112347000-6b7d9800-8cbe-11eb-9169-fc4c714bcb05.mov


core/sass/components/_forms.scss
```diff
@@ -243,3 +243,15 @@
     color: white;
   }
 }
+
+.select-date {
```
dahfool, line 247: Shouldn't this be part of great-styles - https://github.com/uktrade/great-styles/blob/master/src/scss/elements/forms/_general.scss

react-components/src/Helpers.js
```diff
@@ -1,3 +1,23 @@
+const dateFormat = (date) => {
```
dahfool, line 1: Please write some unit test for `dateFormat`, it could return current date if no property is passed to the function.

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -14,6 +14,8 @@ export const Objective = memo(
       })
     }
 
+    const ISONow = new Date().toISOString().slice(0, 10)
```
dahfool, line 17: This function could be move into utils, resusable.

react-components/src/Helpers.js
jakehopking: Added.

core/sass/components/_forms.scss
jakehopking: Moved.

react-components/src/components/ObjectivesList/Objective/index.jsx
jakehopking: Added.

PR #1097, opened 2021-03-24
Fix modals to show on dashboard only

rafa-garcia:
CONTEXT: This changeset fixes a bug where the initial modals were showing on pages beyond the dashboard.

### Merging


core/templates/core/base.html
```diff
@@ -127,14 +127,13 @@
         },
       })
 
-      const loggedInUser = {% if request.user.is_authenticated %}true{%else %}false{% endif %};
-      if (loggedInUser) {
+      {% if tour %}
```
webbyfox, line 130: Am I right, `tour` is restricted to only Logged in User?

rafa-garcia: Indirectly, yes. Tour in the context is available only trough wagtail as a logged-in page

webbyfox: we should have better logic to present the `segment` as I am lost to see how `tour` and `segment` is connected.

webbyfox (approved):
I am approving based on there is some refactoring required as `tour` and `segment` logic need to differentiate.

PR #1100, opened 2021-03-25
GP2-1648: Add in geo-based redirect from GreatDomesticHomePage

Rambowalsh:
(This is a collaboration between Chris and Steve - Steve is writing up this blurb)

This changeset ports Great V1's geo-redirection behaviour to /international/ for users from certain countries:

* if the user comes to the homepage at / from one of four countries (based on their IP) and they haven't been to Magna before, they will be taken to /international/  with their appropriate language code set (Chinese, Spanish, German or Japanese)
* if they return to the homepage after this, they are NOT auto-redirected further because a cookie has been set.

Note that the cookie name has a typo in it: `disable_geoloaction` not `disable_geolocation` but I've deliberately kept it with the typo for now so that it doesn't affect users in the wild who have the typo-ed cookie title. I'll raise this with the PO, too.

(@stevejalim has tested this manually using VPNs and a tunnel and confirms it works as expected)

This changeset retains the City-based behaviour used in great-cms already and adds Country-based lookup from great-domestic-ui.

It also updates the City and Country DBs in the codebase.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1648

### Merging


config/settings.py
```diff
@@ -448,6 +448,10 @@
 
 FEATURE_FLAG_ENABLE_V1_CONTACT_PAGES = env.bool('FEATURE_FLAG_ENABLE_V1_CONTACT_PAGES', False)
 
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+PROJECT_ROOT = os.path.dirname(os.path.abspath(__file__))
+BASE_DIR = os.path.dirname(PROJECT_ROOT)
+
```
stevejalim, line 454: Would be good to move these to the top of the file, by `ROOT_DIR`

core/management/commands/download_geolocation_data.py
```diff
@@ -35,19 +48,37 @@ def retrieve_file(self, edition_id):
...
-        archive.decompress(
-            file_like_object=archive.retrieve_file(edition_id='GeoLite2-City'),
-            file_name=settings.GEOIP_CITY,
-        )
-        archive.decompress(
-            file_like_object=archive.retrieve_file(edition_id='GeoLite2-Country'),
-            file_name=settings.GEOIP_COUNTRY,
-        )
+        geolocation_archive = GeolocationArchiveNegotiator()
+        geolocation_archive.decompress(file_name=settings.GEOIP_COUNTRY, destination=settings.GEOIP_PATH)
```
stevejalim, line 84: Looks like we might need an additional line for settings.GEOIP_CITY too
`geolocation_archive.decompress(file_name=settings.GEOIP_CITY, destination=settings.GEOIP_PATH)`

tests/unit/core/management/commands/test_download_geolocation_data.py
```diff
@@ -21,35 +44,44 @@ def test_handles_remote_invalid_archive(settings, requests_mock):
...
-    assert archive.decompress.call_count == 2
-    assert archive.decompress.call_args_list == [
-        call(file_like_object=city_file, file_name=settings.GEOIP_CITY),
-        call(file_like_object=country_file, file_name=settings.GEOIP_COUNTRY),
-    ]
+    assert geolocation_archive.decompress.call_count == 1
+    assert geolocation_archive.decompress.call_args == call(
+        file_name=settings.GEOIP_COUNTRY,
+        destination=settings.GEOIP_PATH,
+    )
```
stevejalim, line 87: I think we'll also need to expand these tests to cover the GEOIP_CITY database, too

Rambowalsh: Added GeoCity tz file and amrended test_handles_remote_invalid_archive
to run against this new city compressed file

core/management/commands/download_geolocation_data.py
stevejalim: @Rambowalsh  Ah, yeah - looking at existing code in export-plan we still need the GEOIP_CITY stuff to be loaded, so we're gonna have to back-out some of these changes the downloader to make sure we get the country stuff _and_ the city stuff.

stevejalim (changes requested):
heya @Rambowalsh - thanks for this. Looking now more at the surrounding code, I think we need to back out some of the changes to the downloader so that we DO get both the Country and the City data. At the moment it reads like we only now update the country file (which is what i saw when i pulled new data locally, too).

Sorry to bounce this back, but if you could slightly rewind the tests (even by hand, looking at the diff here) that'd be great and would unblock this. Cheers!

stevejalim:
@Rambowalsh Let's check in with each other before you pick this back up, as it might alternatively be better to drop the CITY use and go only with COUNTRY. Will see what I can find out in the meantime

Rambowalsh:
Ok

stevejalim:
@Rambowalsh Yeah, looking around I can see that `core.helpers.get_location()` is used in a couple of places and this `get_location()` function uses the City database, too. So, unfortunately, we'll have to back in the code and tests that load the  city data as well. I think basically anywhere we're loading/fetching/updating the GeoIP Country data we should be doing the same for the City data. Apols for not noticing that earlier in things, but I don't think it's lot of work to add in.

README.md
```diff
@@ -97,6 +97,18 @@ HEADLESS=false make ARGUMENTS="-k test_anonymous_user_should" pytest
  It creates the Great domestic empty homepage and assigns it to the site root.
  It also creates a superuser `test` with password `password`, for local development.
 
+## Geolocation data
+This project includes GeoLite2 data created by MaxMind, available from https://www.maxmind.com/
```
stevejalim, line 101: This was added to comply with the CC attribution requirement noted on https://dev.maxmind.com/geoip/geoip2/geolite2/ @agiamas

stevejalim (commented):
Given this a look over and it seems fine - but I'd rather someone else review it as I've been quite close to the code

webbyfox (approved):
LGTM

PR #1112, opened 2021-03-26
GP2-1915 Align EP data snapshots

bobmeredith:
This PR changes the export plan data snapshots to use the same data mechanisms as WTE and aligns the formatting.
While the WTE page uses a single transparent endpoint that pipes directly though to directory-api, the Export plan has a view that collects all data required for display using a single endpoint on directory-api.  Filtering based on target age groups is also carried out by this view.  Initial population on the Marketing approach page has been removed since this was immediately replaced by an ajax request. The performance hit of double-loading is therefore removed.
Given that the new endpoints are significantly faster than the old ones - a considerable increase in performance is expected from all data-liked EP pages.

Following this work,the entire old population model (which loads population data on-the-fly) can be removed from directory API, along with several other now redundant endpoints.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-1915

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
image
image

core/helpers.py
```diff
@@ -29,29 +30,22 @@
...
-    '65-69',
-    '70-74',
-    '75-79',
-    '80-84',
-    '85-89',
-    '90-94',
-    '95-99',
-    '100+',
-]
+def age_group_mapping(target_ages):
```
depsiatwal, line 33: is it worth worth moving this to directory_constants to make choices consistent with other choices ?

depsiatwal (commented):
Assume it's been tested on changes :-) 

LGTM to me generally

core/helpers.py
bobmeredith: I'm not sure.  The only other place this 'mapping' is used seems to be in javaScript, so it's no use as a python fn.

react-components/src/views/sections/MarketingApproach/index.jsx
```diff
@@ -1,18 +1,25 @@
...
 
+import { Provider } from 'react-redux'
+import Services from '@src/Services'
 import { RouteToMarket } from '@src/components/RouteToMarket'
 import { TargetAgeGroupInsights } from '@src/components/TargetAgeGroupInsights'
 import { FormElements } from '@src/components/FormElements'
 
 export const createRouteToMarket = ({ element, ...params }) => {
-  ReactDOM.render(<RouteToMarket {...params} />, element)
+	ReactDOM.render(<RouteToMarket {...params} />, element)
```
dahfool, line 11: @bob I think it's 2 spacing unit.

depsiatwal (approved):
LGTM

react-components/src/components/TargetAgeGroupInsights/DemoData/index.jsx
```diff
@@ -4,6 +4,15 @@ import { Tooltip } from '@components/tooltip/Tooltip'
 import { Stats } from '@src/components/Stats'
 import { notAvailable } from '@src/components/Stats/StatsGroup'
 import PropTypes from 'prop-types'
+import { millify, normaliseValues, get } from '@src/Helpers'
+
+const formatNumber = (val) => Number(val) ? millify(Number(val)*1000) : notAvailable
+
+const formatLanguages = (languages) => {
```
dahfool, line 11: Could `formatLanguages` in react-components/src/components/TargetAgeGroupInsights/utils.js be used.

bobmeredith: Good thought - I'll do that.

react-components/src/Helpers.js
```diff
@@ -65,6 +65,10 @@ const normaliseValues = (str, places = 1, fixed = false) => {
   return 'Data not available'
 }
 
+const numberWithSign = (value) => {
```
dahfool, line 68: numberWithSign might a default value or return 0 if no params is passed

react-components/src/views/sections/MarketingApproach/index.jsx
bobmeredith: Quite right - and I have no idea why my editor is doing this.  Just seems to be two files.

PR #1125, opened 2021-03-30
GP2-1915 EP AJAX update and remove page reload

bobmeredith:
This changeset removes the need to refresh the screen on a change of product/market for most of the site (particularly the export plan pages)
The data snapshots on export plan pages are now all ajax loaded and  I've removed the data from views to render static information that gets walloped by an immediate AJAX load. There is more tidying that can be done, although the BE views are largely still needed for PDF rendering of the EP. 
I've swapped the logic of the page refresh so that the default is to not refresh - only specially marked pages refresh on a change of product/country. ATM, the only such page is the lesson detail page (as it contains context specific case studies)
One EP page (Marketing approach) has a rather annoying use of country outside any REACT components. This is updated in a slightly bodgy way by the reducer looking for a class in the static elements on the page and updating the name.

### Workflow

### Reviewing help

### Housekeeping

### Merging


react-components/src/actions/index.js
```diff
@@ -12,70 +13,49 @@ export const SET_LOADED = 'SET_LOADED'
 
 // action creators
```
bobmeredith, line 15: These updates are to make eslint a happy bunny

dahfool (approved):

bobmeredith:
Testing
This PR eliminates the page reload on change of product or market.
We need to ensure that all references to target product and market are updated when product/market are changed.  The particular areas of work were the export plan pages with data snapshots on them. Target Markets research, Adaptation for target market, Marketing approach and travel plan.
In addition, Marketing approach has the market name in a title in the bottom half of the page.  Again, this should update correctly.

PR #1127, opened 2021-03-30
fix-failed-pdf-dl-test

depsiatwal:

### Workflow

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/tests/unit/exportplan/test_views.py
+++ b/tests/unit/exportplan/test_views.py
@@ -378,20 +378,22 @@ def test_getting_paid(export_plan_data, client, user, mock_get_user_profile):
 def test_download_export_plan(
     client, mock_get_comtrade_data, mock_get_population_data, mock_cia_world_factbook_data, user, mock_get_user_profile
 ):
-    url = reverse('exportplan:pdf-download')
-    client.force_login(user)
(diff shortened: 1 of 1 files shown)


tests/unit/exportplan/test_views.py
```diff
@@ -375,14 +376,26 @@ def test_getting_paid(export_plan_data, client, user, mock_get_user_profile):
...
+    mock_cia_world_factbook_data,
+    user,
+    mock_get_user_profile,
 ):
+    # Must be a better way of mocking a return object
+    class Errordoc:
+        err = False
+
+    mock_pisa.return_value = Errordoc()
     url = reverse('exportplan:pdf-download')
```
webbyfox, line 394: Just for my understanding, Is this endpoint in Magna? Do we still need to mock it? Thinking it loud that we need a test which should pass valid data to an endpoint which should generate document - Also, we need a bad data test which should validate.

depsiatwal: This is a library context is passed and it generates a message. A context generator already has end point tests for the data using the context class in export plan. Further more the pdf lib can generate an error this flow is also tested.

bobmeredith (approved):
LGTM
We need to get this merged as we have failing unit tests ATM

PR #1149, opened 2021-04-13
Staging release

webbyfox:


stevejalim (approved):
LGTM - i'll add config to enable search after deployment but missing config won't break Staging

PR #1159, opened 2021-04-16
Added url redirect for trade and ISD

webbyfox:
This changeset adds additional redirect from BAU to magna

### DIFF
--- a/tests/unit/config/test_url_redirects.py
+++ b/tests/unit/config/test_url_redirects.py
@@ -468,11 +468,11 @@ def test_redirects_no_trailing_slash(url, expected, client):
         ('/market-research/visit-a-trade-show/', '/advice/find-an-export-market/trade-shows/'),
         (
             '/market-research/doing-business-with-integrity/',
-            '/advice/manage-legal-and-ethical-compliance/understand-business-risks-in-overseas-markets/',  # NOQA
+            'advice/manage-risk-bribery-corruption-and-abuse-human-rights/bribery-and-corruption-understand-risks/',  # NOQA
         ),
         (
             '/market-research/know-the-relevant-legislation/',
-            '/advice/manage-legal-and-ethical-compliance/understand-business-risks-in-overseas-markets/',  # NOQA
+            'advice/manage-risk-bribery-corruption-and-abuse-human-rights/bribery-and-corruption-understand-risks/',  # NOQA
         ),
         ('/business-planning/', '/advice/define-route-to-market/'),
         ('/business-planning/make-an-export-plan/', '/advice/create-an-export-plan/how-to-create-an-export-plan/'),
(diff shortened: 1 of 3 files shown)


tests/unit/config/test_url_redirects.py
```diff
@@ -620,8 +620,12 @@ def test_redirects_no_trailing_slash(url, expected, client):
...
         ),
+        ('/trade/', '/international/trade/incoming/'),
         ('/trade/some-term/', '/international/trade/incoming/some-term/'),
+        ('/investment-support-directory/', '/international/investment-support-directory/'),
         ('/investment-support-directory/some-term/', '/international/investment-support-directory/some-term/'),
+        ('/story/york-bag-retailer-goes-global-via-e-commerce/', '/success-stories/york-bag-retailer-goes-global/'),
+        ('/story/hello-babys-rapid-online-growth/', '/success-stories/hello-babys-rapid-online-growth/'),
     ],
 )
 def redirect_articles(incoming_url, expected_url, client):
```
webbyfox, line 631: @stevejalim do you know why this method is not prefix/suffix with `test_`?

webbyfox: look like wasn't running as part of testsuits

stevejalim: You're right - and wasn't running in V1 either - good catch!

tests/unit/config/test_url_redirects.py
```diff
@@ -468,11 +468,11 @@ def test_redirects_no_trailing_slash(url, expected, client):
         ('/market-research/visit-a-trade-show/', '/advice/find-an-export-market/trade-shows/'),
         (
             '/market-research/doing-business-with-integrity/',
-            '/advice/manage-legal-and-ethical-compliance/understand-business-risks-in-overseas-markets/',  # NOQA
+            '/advice/manage-legal-and-ethical-compliance/understand-business-risk-in-overseas-markets/',  # NOQA
```
stevejalim, line 471: Both this edited version and the original (risks and risk) both 404 in production -- I guess that means this is a dead redirect anyway? Maybe worth mentioning on the ticket for Jess to confirm if there's a different destination they want

tests/unit/config/test_url_redirects.py
```diff
@@ -468,11 +468,11 @@ def test_redirects_no_trailing_slash(url, expected, client):
...
         ),
         (
             '/market-research/know-the-relevant-legislation/',
-            '/advice/manage-legal-and-ethical-compliance/understand-business-risks-in-overseas-markets/',  # NOQA
+            '/advice/manage-legal-and-ethical-compliance/understand-business-risk-in-overseas-markets/',  # NOQA
```
stevejalim, line 475: Same as above

tests/unit/config/test_url_redirects.py
```diff
@@ -521,12 +521,12 @@ def test_redirects_no_trailing_slash(url, expected, client):
             '/advice/manage-payment-for-export-orders/payment-methods-for-exporters/',  # NOQA
         ),
         (
-            '/getting-paid/consider-how-youll-get-paid/',
+            '/getting-paid/consider-how-to-get-paid/',
```
stevejalim, line 524: Nice fix/catch.

stevejalim (approved):
Looks good to me - thanks! One question raised re a URL you had to change -- might be worth checking with Jess about that one at Test stage

tests/unit/config/test_url_redirects.py
webbyfox: will do... I just check existing definition on code and amended accordingly

PR #1176, opened 2021-04-21
GP2-2137 VFM companies house lookup

bobmeredith:
Adds 'company name' pages to the VFM survey.
Users have a type-ahead that allows them to begin typing the company name. A lookup is carried out (debounced at 250ms) and the results presented in a drop-down.
This PR piggy-backs off the custom 'Select' component we already have. Just a few extra props were added to create and populate a text input within the Select. A few styling tweaks were needed to push the drop-down below the input and to remove the placeholder line.
There are several 'pages' with the company name question. Once the user selects a company, modal is populated with a review of the selected company using data drawn from both the search result and another request to 'get_profile' from CH.
Alternatively, the user can switch to an option to simply type the company name - no CH integration is required for this.

The CH lookup uses our CH interface server, so the url and key will need to be populated.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2137

### Reviewing help

### Housekeeping

- [x] Cleaned up old feature flags

### Merging


bobmeredith:
image

image

image

image

image

bobmeredith:
Testing:
Note - if working locally, will need a local instance of company house API webserver.
Trigger VFM questionnaire and work through to the company name question.
Try navigating forward and back to company.
Check company details.
In directory-sso, check that the sic codes are being saved against company.
Check that the manual entry mode works, that it saves only the company name and that navigation to and from restores to the correct search mode.

dahfool (approved):

PR #1197, opened 2021-04-26
GP2-2176 VFM Multiselect

bobmeredith:
To get the Select component working in multi-select mode for VFM, I needed to expse the values from the options rather than labels.
So, this changeset converts the select itself to always deal with option values rather than labels, or, in the case of a multi-select, an array of values.
This change required a considerable re-work around export plan, but is much simpler as the export plan itself stores values and arrays of values rather than labels. Four helper functions that mapped the select label-based output can be removed and externally, the behavior of the select in single or multi becomes much simpler.

Also, added a keydown handler on the autocomplete input with a test.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2176

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
Testing:
All EP sections that contain <select> input should be manually tested.

- Check the initial state with a new EP.
- Check that selects and multi-selects can be changed.
- Check that the change stays when the page is refreshed

dahfool (approved):

PR #1284, opened 2021-05-18
GP2-2721: Add support for subtitles on GreatMedia videos

stevejalim:
This changeset adds field to GreatMedia for English subtitles (for now - more to come) which are then
served to the frontend via a dedicated view. The regular HTML5 video player picks up these subtitles
via a dedicated <track> node per language.

Note that VTT subtitles are required, and if the subtitle format is at all incorrect, things will not load.
This is partly why the subtitles are stored flat in a textfield, rather than being in a separate file: it's
much easier to tweak/tune them this way.

By default, no subtitles are shown - they have to be manually enabled.

Final note: subtitles are currently put behind a login_required decorator, which means they'll only work for pages behind authentication. We could relax this in the future.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2721

### Reviewing help

Screenshot 2021-05-18 at 13 01 50

Chrome
Screenshot 2021-05-18 at 11 13 33

Safari
Screenshot 2021-05-18 at 11 07 07

Firefox
Screenshot 2021-05-18 at 11 06 35

### Merging


core/urls.py
```diff
@@ -109,5 +110,15 @@ def anonymous_user_required(function):
         name='api-trade-barrier-data',
     ),
     path('api/companies-house/', skip_ga360(views_api.CompaniesHouseAPIView.as_view()), name='api-companies-house'),
+    path(
+        'subtitles/<int:great_media_id>/<str:language>/content.vtt',
+        login_required(
+            # NB remove/update login_required() if we start serving subtitles for videos in
+            # unauthed pages, but then ideally move to a UUID for the GreatMedia instance
```
webbyfox, line 117: Steve, Can you please confirm that all videos are behind auth atm?

webbyfox (approved):
LGTM - Just small clarification question

stevejalim:
@webbyfox 

The video files themselves are in a publicly-readable bucket (but not one with a public index to click through). 
The lesson videos are only used in pages which require auth (the lessons).
 
Videos on unauthenticated pages (eg in /trade-finance/) do not use the same templatetag to render them as the lessons, and so do not get subtitles. Adding subtitles support for pages that can be viewed without auth would be a need us to remove login_required from the subtitles view, and ideally refactor the URL as per our chat before

agiamas (approved):
lgtm

PR #1292, opened 2021-05-19
Update README.md with MacOS Big Sur setup addition

stevejalim:
openssl is needed for the LDFLAGS advice to work, but we weren't making that clear


depsiatwal (approved):
LGTM

PR #1334, opened 2021-05-24
enable-apm-logging

depsiatwal:

### Workflow

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -363,6 +363,18 @@ if DEBUG:
     MIDDLEWARE = ['debug_toolbar.middleware.DebugToolbarMiddleware'] + MIDDLEWARE
     INTERNAL_IPS = ['127.0.0.1', '10.0.2.2']
 
+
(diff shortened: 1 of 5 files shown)


stevejalim (approved):
Thought for a moment about `default=not DEBUG`, but I think it won't trip us up in real use. 👍

ejayesh (changes requested):
Can we remove grea-cms service name, we may change need to change the at some point, better to change a VAR the code.

config/settings.py
```diff
@@ -363,6 +363,18 @@
     MIDDLEWARE = ['debug_toolbar.middleware.DebugToolbarMiddleware'] + MIDDLEWARE
     INTERNAL_IPS = ['127.0.0.1', '10.0.2.2']
 
+
+ELASTIC_APM_ENABLED = env("ELASTIC_APM_ENABLED", default=not DEBUG)
```
stevejalim, line 367: let's give this an explicit default and then set `config/dev` to `False` for it, rather than `default=not DEBUG`, so that tests don't get tripped up

PR #1359, opened 2021-06-03
GP2-2879 DAC_Parsing_01 (A)

Zak-JS:
Issue with incorrect usage of li and ul tags in reference to accesibliity.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2879

### Reviewing help

### Housekeeping

### Merging


bobmeredith (commented):
Looking at the audit, there is more still to do on this. They suggest an <ol> in each section and <li>s inside that surrounding the spans that surround the buttons.
I'd say the extra <li> is not needed and you could just convert the span to an <li> with the right classes.

bobmeredith (approved):
LGTM

PR #1402, opened 2021-06-15
GP2-2982: Rebuild sitemap.xml in great-cms

stevejalim:
This changeset adds a hybrid sitemap to great-cms, so that we no longer have to use the older one routed from great-domestic-ui. It is "hybrid" because it shows all published, public/no-auth-requrired Wagtail pages, as well as all Django-rendered views that have been migrated from great-domestic-ui so far. (It has some hard codig of /contact/* pages, because those are still to be ported from great-domestic-ui.)

Note that neither the original sitemap nor this new one make any reference to pages within great-international-ui.

There have been some changes to the sitemap. (I'll be asking an SEO specialist within DIT to validate that these changes make sense and you can check/compare what's below with the currently live sitemap at https://www.great.gov.uk/sitemap.xml)

* the old sitemap loaded in all possible URLs for GDUI and then excluded some. This meant that the sitemap included URLs didn't make much sense to suggest to search engines, such as 'success' pages that are shown after some forms are submitted, as well as technical/internal-use paths like /healthcheck/. This kind of thing is no longer done -- the sitemap only references pages that make sense to promote to search engines.
* articles (eg advice) were deliberately excluded from the sitemap, for some reason, but are now included
* market guides, similarly, were also excluded but are now included
* the "changefreq" value of any Django-rendered page used to be 'daily' but i've set this to 'monthly', given how rarely these pages change. We can always request a re-crawl if things are particularly important to get re-indexed.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2982

### Reviewing help

* locally, got to /sitemap.xml -- the content will depend on what pages are in your local CMS

### Merging


depsiatwal (approved):
LGTM

webbyfox (approved):
LGTM

PR #1420, opened 2021-06-22
GP2-3065 Simplify 3CE url config

bobmeredith:
3CE changed the base url for our development use, (from dev to staging) and this is making the current env-var config unwieldy. 

It is now necessary to configure three long urls (plus one token) in all environments including dev's local environments.
If we change to using a base url (but keeping the potential to configure individual service urls if needed), we could allow devs and the non-prod environments to require only the token to be configured, while the prod environment would require only the prod token plus a single prod base url.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3065

### Reviewing help

### Housekeeping

- [ ] Cleaned up old feature flags (can't do this until new code rolled out)

### Merging


config/settings.py
```diff
@@ -521,16 +521,15 @@
 
 
 # 3CE commodity classification
+CCCE_BASE_URL = env.str('CCCE_BASE_URL', 'https://info.stage.3ceonline.com')
 COMMODITY_SEARCH_URL = env.str(
-    'CCCE_COMMODITY_SEARCH_URL', 'http://info.dev.3ceonline.com/ccce/apis/classify/v1/interactive/classify-start'
+    'CCCE_COMMODITY_SEARCH_URL', CCCE_BASE_URL + '/ccce/apis/classify/v1/interactive/classify-start'
 )
 COMMODITY_SEARCH_REFINE_URL = env.str(
```
webbyfox, line 528: Do we need `COMMODITY_SEARCH_URL` and `COMMODITY_SEARCH_REFINE_URL` as env vars then? Only changes will be base URL, right?

bobmeredith: Yes it would. 
I was just leaving the potential to change those individual urls if 3CE decide to mix things up. Plus, support of those urls means it's backwards compatible - as all those individual urls are currently set up in all the envs, so we could deploy to prod without changing anything immediately.

bobmeredith: Actually, I think you're right. Simplicity trumps backward compatibility given that local instances should be on staging so everything apart from production would work without any changes to config. I'll amend.

webbyfox (approved):

PR #1434, opened 2021-07-02
GP2-2988: Fix SOO contact form - make empty name fields editable

stevejalim:
BUGFIX: Make the name fields in the SOO contact wizard to still be editable if the first and/or last name is not yet known

This is needed because Great V2 allows users to sign up without providing a first or last name, which would then cause the overall wizard to 500.

**NOTE THAT this fix is being applied to the pages after they have been migrated from V1 into V2, but if we decide we want to hotfix this to production on GDUI, it is a copy-paste change to port over.**

Tests work at the Python level, but tested manually, too, of course.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2988

### Reviewing help

- [X] Includes screenshot(s)

**BEFORE** - see https://uktrade.atlassian.net/wiki/spaces/Great/pages/2681405487/SOO+applications+-+make+first+and+last+name+fields+editable

**AFTER**

Screenshot 2021-07-01 at 18 04 26
Screenshot 2021-07-01 at 18 04 15

### Merging

- Leave to Steve to merge into a different target branch, after GP2-1648 has been merged to the contact-pages feature branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Enhancements
 - GP2-1684: Port the contact pages which need users to be signed in (SOO ones) from V1 to V2
 ### Fixed bugs
-
+- GP2-2988: Make SOO contact form name fields editable if not already populated.
 ## [1.11.0](https://github.com/uktrade/great-cms/releases/tag/1.11.0)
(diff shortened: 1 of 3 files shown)


webbyfox (approved):
welldone @stevejalim, Nice one

PR #1438, opened 2021-07-06
GP2-3071 fix Docker SSO hostnames

froddd:
Specify correct SSO hostnames in  for running all services in Docker.

Moves secrets env vars loading last: all local env vars can then be set in secrets file, which runs no risk of committing them to the repo.

### Workflow

### Reviewing help

### Housekeeping

### Merging


depsiatwal (approved):
LGTM
but do we need them are they not picked up from 'dev'

froddd:
The relevant vars in `dev` are incorrect for running in Docker (hostname does need to be correctly specified as `sso.proxy.trade.great` instead of current `sso.trade.great`). These could be updated, but it may be a breaking change as it would require anyone running Great CMS locally to be aware they need to update their hosts file to include:
```
127.0.0.1    sso.proxy.trade.great
```
Happy to go ahead with that; this would be the cleaner way of doing things.

depsiatwal (approved):
LGTM

PR #1453, opened 2021-07-15
GP2-2492 Export plan - some pages - value input

seijihg:
The numbers will be invisible as soons as user click on number fields.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2942

### Reviewing help

- [] Includes screenshot(s) - ideally before and after, but at least after

### Housekeeping

### Merging


bobmeredith (changes requested):
there is only a change to scss, so there really shouldn't be a rebuild of js in here.
You sure you did an npm build, rather than adding the files from the watch

seijihg:
> there is only a change to scss, so there really shouldn't be a rebuild of js in here.
> You sure you did an npm build, rather than adding the files from the watch

npm run build done! thanks

bobmeredith (approved):
LGTM

PR #1470, opened 2021-07-27
GP2-2896 split company objectives dates

froddd:
Split company objective dates into month and year.

Related to https://github.com/uktrade/directory-api/pull/924

UI updated to match designs:
Screenshot 2021-07-27 at 12 07 01

Serializers also updated to match new format.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2896

### Reviewing help

### Housekeeping

### Merging


depsiatwal (approved):
LGTM

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -6,101 +6,154 @@ import { Input } from '@src/components/Form/Input'
...
+            id="description"
+            placeholder="Add some text"
+            label={`Objective ${number}`}
+            value={data.description}
+            onChange={onChange}
+            errors={[]}
+            formGroupClassName="m-b-0"
+          />
+        </div>
+        <div className="costs__option costs__option--border">
```
bobmeredith, line 70: Consider building a little component for a 'date' in this format. I guess, at some stage could extend to add a 'format' prop where this month+year format is just one of the possibilities.
Anyway - it would dry up.

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -6,101 +6,154 @@ import { Input } from '@src/components/Form/Input'
...
+                  label="Month"
+                  id="start_month"
+                  name="start_month"
+                  update={onChange}
+                  options={selectMonths}
+                  selected={`${data.start_month}`}
+                />
+              </div>
+              <div className="inputgroup__input inputgroup__input--year">
                 <Input
```
bobmeredith, line 85: Just wondering - maybe one to discuss with Michal. Why not use a pair of selectors instead of one selector and one input? 
The nature of this date entry is that the year is only useful to be or in the future and I guess 10 years would be plenty (like any company looks that far ahead)
Advantages - a) Much nicer for users not to have to switch from clicking/touching to number entry - particularly on mobile.
b) Possibly easier to align and code as we have two of the same controls.

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -116,8 +169,10 @@ Objective.propTypes = {
     description: PropTypes.string,
     owner: PropTypes.string,
     planned_reviews: PropTypes.string,
-    start_date: PropTypes.string,
```
bobmeredith, line 119: TBH, I think I would have stuck with storing the date as a date rather than in separate fields - just set the date to 1 perhaps. It could lead to issues with trying to save data partway through entry, but gives an easier way back if the designers change their mind and seems neater in the BE.

react-components/src/components/ObjectivesList/Objective/index.jsx
froddd: Absolutely -- I need it for my next ticket already!

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -6,101 +6,154 @@ import { Input } from '@src/components/Form/Input'
 import { ConfirmModal } from '@src/components/ConfirmModal/ConfirmModal'
 import { objectHasValue } from '@src/Helpers'
 import ErrorList from '../../ErrorList'
+import { Select } from '@src/components/Form/Select'
 
+const MONTHS = [
```
bobmeredith, line 11: I can't help feeling these should be available inside a Date object somewhere - but I guess literals it is.

react-components/src/components/ObjectivesList/Objective/index.jsx
froddd: As discussed, will use input for now -- I've adjusted pattern so it brings numeric keypad up on iOS.

react-components/src/components/ObjectivesList/Objective/index.jsx
froddd: original dates are left in place so this migration is non-destructive, and can be reverted!

react-components/src/components/ObjectivesList/Objective/index.jsx
froddd: As discussed, will use input for now -- I've adjusted pattern so it brings numeric keypad up on iOS.

froddd: As discussed, will use input for now -- I've adjusted pattern so it brings numeric keypad up on iOS.

react-components/src/components/ObjectivesList/Objective/index.jsx
froddd: original dates are left in place so this migration is non-destructive, and can be reverted!

bobmeredith (approved):
LGTM

PR #1471, opened 2021-07-27
GP2-2893 back-link placement

bobmeredith:
Fixes issues with placement of backlinks from learning detail and curated list pages.
Also fixes the issue with the text dropping to a very narrow column on mobile.
The column issue is casued by the fact that at the mobile breakpoint, columns all go 100% width (on the basis that they should stack) On this page though, it means the centre column is squashed to 1/3 width rather than the 1/2 it should be. After experimenting with hiding the side columns on mobile, the most satisfactory solution with minimum transitional jarring seemed to be to put a min width on the centre column so it gradually takes more of the screen as the width collapses.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-2893

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
image

image

image

core/sass/learn/_base.scss
```diff
@@ -375,8 +375,8 @@ $desktop: 1024px;
 }
 
 .learn__back-link {
-  left: 38px;
-  top: 15px;
+  left: 0px;
```
froddd, line 378: It seems odd to even want the link positioned absolutely on mobile. Would it make more sense to set the link inside the container and allow natural document flow? If it needs to be positioned left for larger screens then use the grid as intended, i.e. with the link in its own `.c-1-4` column?

froddd (approved):
LGTM

core/sass/learn/_base.scss
bobmeredith: Possibly, but the link is perhaps longer than the LH column. I'm not completely sure of the rationale behind absolute positioning, but losing that seems to give a world of pain.
The link actually hides on mobile.  That was the bug in fact - the hide-on-mobile class didn't work because there was scss on this el that set display to 'flex'. All I did was to move the 'hide' to the outer span rather than the <a> with the flex on it.

PR #1481, opened 2021-08-05
GP2-3119 Decouple basket

bobmeredith:
Decouples the users products and markets from the export plan.
the user's markets and products are now saved in user_data in directory sso and are pulled in by ajax rather than being baked into any pages.
The basket has a new item, in global state - in addition to the EP plan one. The EP global state item is still present and will be pre-loaded with the EP in context on EP pages. The basket item may be pre-loaded, but ATM is only populated by an ajax call when required. Basket updates are through the reducer. 

### Workflow

### Reviewing help

### Housekeeping

### Merging


react-components/src/components/ProductFinder/BasketView.jsx
```diff
@@ -0,0 +1,57 @@
...
+  },
+  content: {},
+}
+
+export default function BasketViewer({ label, onOpen, children }) {
+  const [modalIsOpen, setIsOpen] = useState(false)
+  const buttonRef = useRef(null)
+
+  const openViewer = () => {
+    const { left, top, height } = buttonRef.current.getBoundingClientRect()
```
froddd, line 21: Can the positioning be solved with CSS? `getBoundingClientRect` doesn't play nice with tests... Ever

react-components/src/components/ProductFinder/CountryFinderButton.jsx
```diff
@@ -66,41 +80,19 @@ CountryFinderButton.propTypes = {
...
-      dispatch(actions.setMarket(market))
-    },
-  }
-}
-
-const ConnectedContainer = connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(CountryFinderButton)
-
```
froddd, line 96: Hallelujah! 🙌

react-components/src/components/ProductFinder/CountryFinderButton.jsx
```diff
@@ -2,59 +2,73 @@ import React, { useState } from 'react'
...
-  const openModal = () => {
-    setConfirmationRequired(!!market)
-    setIsOpen(!market)
+  const loadUserMarkets = () => {
+    // Load the market list into redux.  We'll be needing it later
+    if (!markets)
+      Services.getUserData(userMarketsKey).then((result) => {
+        Services.store.dispatch(actions.setMarkets(result[userMarketsKey]))
+      })
+  }
```
froddd, line 29: Seen this somewhere else -- could this be turned into a hook? It's probably not a huge performance penalty to run it on every page load, and I assume it may do anyway?

react-components/src/components/ProductFinder/CountryFinderModal.jsx
```diff
@@ -92,6 +92,12 @@ export default function CountryFinderModal(props) {
     })
   }
 
+  useEffect(() => {
+    if(commodityCode) {
```
froddd, line 96: This seems redundant as `getSuggestedCountries` already checks for this

core/templates/core/includes/personalisation_bar.html
```diff
@@ -1,20 +1,16 @@
 
 {% load wagtailimages_tags %}
 {% load routablepageurl from wagtailroutablepage_tags %}
-{% with product=request.user.export_plan.data.export_commodity_codes.0 market=request.user.export_plan.data.export_countries.0 %}
 <div class="grid bg-blue-deep-80 shared-personalisation-bar">
-  <nav class="m-f-s p-f-0 p-r-xs p-v-xs c-full-width text-white container shared-personalisation-bar-nav" id="personalisation-bar">
-      <span class="m-r-xxs">I want to export</span>
+  <nav class="m-f-s p-f-0 p-r-xs p-v-xxs c-full-width text-white container shared-personalisation-bar-nav" id="personalisation-bar">
```
froddd, line 5: Unrelated to current work: why is this even a grid? 🤔

react-components/src/components/DataSnapShot/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { memo } from 'react'
 import PropTypes from 'prop-types'
 import { useSelector } from 'react-redux'
-import { getProducts, getMarkets } from '@src/reducers'
+import { getEpProduct, getEpMarket } from '@src/reducers'
```
froddd, line 4: If we're keeping products and markets relations in EP it might make sense to keep these method names as plural to avoid having to change again/confusion later (makes this change ever so slightly smaller in some ways, too)

react-components/src/components/ProductFinder/ProductFinder.test.jsx
```diff
@@ -182,53 +215,49 @@ describe('Product finder tests', () => {
...
       expect(results).toBeTruthy()
     })
-    expect(finder.querySelector('.form-group-error').textContent).toMatch(/We couldn't find a match because your search had too many product names/)
+    expect(finder.querySelector('.form-group-error').textContent).toMatch(
+      /We couldn't find a match because your search had too many product names/
+    )
   })
 
-  it('Opens product view and renames product', async () => {
+  xit('Opens product view and renames product', async () => {
```
froddd, line 243: Just a note about this, as other skipped test -- would definitely benefit from being run/fixed!

react-components/src/components/ProductFinder/ProductFinderButton.jsx
```diff
@@ -1,72 +1,70 @@
...
 import { getProducts } from '@src/reducers'
-import { connect, Provider } from 'react-redux'
+import { Provider, useSelector } from 'react-redux'
 import Services from '@src/Services'
+import actions from '@src/actions'
 
 import ProductFinderModal from './ProductFinderModal'
+import BasketViewer from './BasketView'
 
+const userProductsKey = 'UserProducts'
```
froddd, line 13: Is there a central place all these could be defined?

react-components/src/components/ProductFinder/ProductFinderButton.jsx
```diff
@@ -1,72 +1,70 @@
...
   const [modalIsOpen, setIsOpen] = useState(false)
+  const selectedProducts = useSelector((state) => getProducts(state))
+
+  const loadProducts = () => {
+    // Load the product list into redux.  We'll be needing it later
+    if (!selectedProducts)
+      Services.getUserData(userProductsKey).then((result) => {
+        Services.store.dispatch(actions.setProducts(result[userProductsKey]))
+      })
+  }
```
froddd, line 25: This also seems repeated (or likely to be) -- hook?

react-components/src/components/ProductFinder/ProductFinderModal.jsx
```diff
@@ -36,18 +36,23 @@ export default function ProductFinderModal(props) {
...
     setSearchResults()
-    if (onCloseRedirect && !selectedProduct) {
-      window.location.href = onCloseRedirect
-    }
   }
 
   const modalAfterOpen = () => {
     setIsScrolled({})
     setSearchTerm('')
+    loadProducts()
```
froddd, line 55: As above, though I can see the use case, only fetching these when the modal opens. Would be a handy content hook if modal contents and functionality were separated for all these components.

react-components/src/components/ProductFinder/ProductFinderModal.jsx
```diff
@@ -452,7 +459,7 @@ export default function ProductFinderModal(props) {
             onClick={search}
             aria-label="search item"
           >
-            <inline className="visually-hidden">Search product</inline>
```
froddd, line 455: What even was that?! Nice!

core/templates/core/base.html
```diff
@@ -109,9 +109,9 @@
           login: false,
           signup: false,
         },
-        exportPlan:{
-          products: {{request.user.export_plan.data.export_commodity_codes|to_json}},
-          markets: {{request.user.export_plan.data.export_countries|to_json }}
+       exportPlan:{
```
froddd, line 112: Slight misalign here

bobmeredith:
Testing - expect it to be broken!
Seriously - the personalisation bar should be replaced by a pair of drop-down type thingies that list the 'basket' items and allow you to add via the same'ol modals. The review/rename feature on products is not present in the design any more.
On WTE, the product list is presented as a <select> on the product tab.
Note that the basket is not baked into every page, but loaded as needed - hence a short delay on opening the drop-down. Changes to the basket (add and delete) should reflect on the WTE select immediately.
In the export plan, the data snapshots should work as before but with the product and market from the EP.

bobmeredith:
image

image

froddd (commented):
Looks good (and huge!), just added a few comments. I may not have got my head correctly round some of the patterns so excuse me if some of that's the case for a few of these!

react-components/src/components/ProductFinder/BasketView.jsx
froddd: I see -- this is so the modal displays in the correct place as the personalisation bar is scrollable... Not ideal! Does it need to be scrollable? Seems like the dropdowns should appear in the flow instead of in the Modal: opening the Markets basket on mobile hides half of it.

react-components/src/components/ProductFinder/BasketView.jsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  const closeModal = () => {
+    setIsOpen(false)
+  }
+
+  const triggerButton = (
+    <button
+      type="button"
+      className={`tag ${modalIsOpen ? 'tag--tertiary' : 'tag--secondary'}`}
```
froddd, line 34: needs extra `tag--icon` class to position and size icon correctly

bobmeredith: Ah, very true. Thanks.

react-components/src/components/ProductFinder/CountryFinderButton.jsx
bobmeredith: Absolutely. As it happens, the issue loading up from an empty DB needs a fix that I don't want duplicated.

react-components/src/components/ProductFinder/CountryFinder.test.jsx
```diff
@@ -55,7 +55,7 @@ it('Opens and closes country finder', async () => {
   expect(document.body.querySelector('.country-finder')).toBeFalsy()
 })
 
-it('Opens with confirmation', async () => {
+xit('Opens with confirmation', async () => {
```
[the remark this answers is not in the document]
bobmeredith: Nah, I've actually deleted the test. This functionality is no longer required or implemented

react-components/src/components/ProductFinder/ProductFinderButton.jsx
bobmeredith: Indeed - it's all refactored now.

react-components/src/components/ProductFinder/ProductFinderModal.jsx
bobmeredith: IMHO, the personalisation bar should be deleted completely - but the designers are unable to see it, At least, it's pretty much useless in most pages.
Hence, I wanted to avoid the overhead of populating it.  Not even worth the BE effort of retrieving the data and baking into every page on the off chance that it may be used.
I managed to get the 'designers' to drop the count of items from the bar, so it's fine to load only when a drop-down is opened.

core/templates/core/base.html
bobmeredith: Now that's careful reviewing!  Thanks.

react-components/src/components/ProductFinder/ProductFinder.test.jsx
bobmeredith: I added a comment in the code.
The rename functionality is no longer reachable in the current design, but I assuming that the designers will see sense at some stage and put it back - hence, I didn't want to delete the test just yet.

react-components/src/components/ProductFinder/BasketView.jsx
bobmeredith: So, ATM there is a problem in that the personalisation bar is located in the header which has hidden overflow - and in a scrollable section as you spotted.
The need for the section to be scrollable is probably no longer valid. It was to allow for very long product names - but now the product name no longer shows in the bar.
I'll look into how things can be re-arranged, but I'm bit nervous that we could easily get into a game of whack-a-mole with layout.

core/templates/core/includes/personalisation_bar.html
bobmeredith: Can take a look at this. Many of the requirements of the existing personalisation bar have now gone, so it's possibly up for a restyle.

react-components/src/components/ProductFinder/ProductFinderButton.jsx
bobmeredith: You're quite right. I knew this was repetitive, but couldn't immediately see a way to centralize.
So now, I've created a hook that handles the whole read/write and store of user variables, and exposed curried versions for each of the settings. loads of code can go including pointless redux boilerplate.

react-components/src/components/ProductFinder/BasketView.jsx
froddd: Might make sense to follow other modals and make that full screen with a close button for mobile?

bobmeredith: Possibly, but that's not the design.
IMHO, the whole thing should be deleted.

react-components/src/components/CompareMarkets/SocietyTab.test.jsx
```diff
@@ -96,19 +96,20 @@ describe('Compare markets - Society tab', () => {
...
       apiCountryDataUrl: '/api/data-service/countrydata/',
+      apiUserDataUrl: '/sso/api/user-data/-name-/',
       user: { id: '6' },
     })
     fetchMock.get(/\/export-plan\/api\/society-data\//, societyApiResponse)
     fetchMock.get(/\/api\/data-service\/countrydata\//, countryDataApiResponse)
-    fetchMock.get(/\/sso\/api\/user-data\//, () => comparisonMarketResponse)
+    fetchMock.get(/\/sso\/api\/user-data\/ComparisonMarkets\//,() => {
+      return comparisonMarketResponse
+    })
```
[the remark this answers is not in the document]
bobmeredith: It was being updated, but no more, so yup.

react-components/src/components/ProductFinder/CountryFinderModal.jsx
bobmeredith: True.  Updated

PR #1491, opened 2021-08-12
GP2-3119 Basket restyle

bobmeredith:
Re-style of the persionalisation bar basket drop-downs so they no longer use react-modal.
They are now absolute positioned with a click/focus catcher to close.

Also a tidy-up of the p-bar declaration to remove the now unused product/market from the export plan.
Fix to product selector on product tab so that it works if more than one product share a single HS6 code. 

### Workflow

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
image

react-components/src/components/ProductFinder/BasketView.jsx
```diff
@@ -1,38 +1,32 @@
...
+    if (!modalIsOpen) {
+      onOpen()
+    }
   }
 
-  const closeModal = () => {
-    setIsOpen(false)
-  }
+  useOnOutsideClick(outerSpan, (target) => {
+    if(!target.closest('.ReactModalPortal')) {
```
froddd, line 18: Is this targeting the right element?

bobmeredith: useOnOutsideClick will check for the element being clicked (or focussed) to be a descendant of the element originally passed to it (the container of the basket drop-down in this case) and will squash such clicks. 
In this case though, If the user opens the product or country finder modal from the basket, it closes as focus transfers inside the modal. 
I think the UX is better if the basket doesn't close in this case - as the user will see the list of products/markets extend.
useOnOutsideClick passes the clicked or focussed target, hence I'm checking if it's a child of a react-modal. It might seem a bit vague, but the only way to open a modal when the basket is expanded would be to click on the button inside the basket - otherwise there would be a prior event that would have already closed this drop-down.

froddd: Aaaaah yes of course -- I read that a bit too quick. Thought it was odd to still reference  the modal in a PR about removing the modal. Should have rubber-ducked! Nice detail.

froddd (approved):

PR #1508, opened 2021-08-18
GP2-3271, GP2-3291, GP2-3119, GP2-3168, GP2-3166, GP2-3292, GP2-3293, GP2-3294, Styling fixes and basket delete

bobmeredith:
A couple of minor styling fixes plus a fix for deleting from basket (so it gets the right product/market)
The issue was that when I introduced sorting, the delete continued to use the index of the item as displayed.
Switching the display to use a sort map allows it to retain the original index.

### Workflow

No changelog since these issues are on in-flight tickets

### Reviewing help

### Housekeeping

### Merging


react-components/src/components/ProductFinder/CountryFinderButton.jsx
```diff
@@ -21,34 +22,39 @@ export const CountryFinderButton = () => {
...
-            </li>
-          ))}
+          {sortMap.map((marketIdx) => {
+            const market = markets[marketIdx]
+            return (
+              <li className="p-v-xxs" key={`market-${marketIdx}`}>
+                <button
+                  type="button"
+                  className="f-r button button--small button--only-icon button--tertiary"
+                  onClick={() => deleteMarket(marketIdx)}
```
froddd, line 40: I can see that this works, but it seems to introduce a third-party index in the process, in the form of that 'sorted map' -- could `deleteMarket` take the `country_iso2_code` instead, removing the need to pass the 'sorted map' around?

bobmeredith: Indeed, I could use the iso code for markets as they are unique.  The same approach would not however, work for products as they have no unique identifier.
I wanted the two lists to use the same mechanism as I'm feeling that there is a lot of commonality between the two and it might be possible to dry up some of the functionality at some stage.

froddd: Yup, that makes sense. Happy to go with that!

froddd (approved):

PR #1526, opened 2021-08-31
GP2-3283 Delete export plan

bobmeredith:
Delete export plan.
Adds a button on the EP homepage to delete just that one plan.
There is a confirmation modal when the delete button is clicked to prevent accidental deletion.  Because of this, it's a largely React implementation, calling an Ajax endpoint to do the deletion and then jumping to the EP index page if successful.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3283

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
image

image

exportplan/templates/exportplan/dashboard_page.html
```diff
@@ -67,10 +67,12 @@ <h3 class="m-v-xs bold">Upload your logo</h3>
     </div>
     {% endif %}
     {% with export_plan_id=export_plan.pk %}
-    <a class="button button--secondary button--full-width button--icon m-b-xs" href="{{export_plan_download_link}}" title="Save your plan as a PDF">
+    <a class="button button--secondary button--small button--full-width button--icon m-b-xs" href="{{export_plan_download_link}}" title="Save your plan as a PDF">
```
froddd, line 70: `title` attribute needs updating to new wording

exportplan/core/helpers.py
```diff
@@ -13,6 +13,12 @@ def create_export_plan(sso_session_id, data):
     return response.json()
 
 
+def delete_export_plan(sso_session_id, id):
+    response = api_client.exportplan.delete_export_plan(sso_session_id=sso_session_id, id=id)
```
froddd, line 17: I take it the `delete_export_plan` methods makes sure the export plan ID belongs to the session user? Just checking ;)

react-components/src/components/Dashboard/DeleteButton.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+      const { queryByText, getByText } = setup()
+      const button = getByText('Delete plan')
+      expect(button).toBeInTheDocument()
+      fireEvent.click(button)
+      expect(getByText(confirmationTitle)).toBeInTheDocument()
+      fireEvent.click(getByText('Cancel'))
+      expect(queryByText(confirmationTitle)).toBeFalsy()
+    })
+
+    xit('Should delete export plan', () => {
```
froddd, line 38: Should this be skipped?

react-components/src/views/sections/Dashboard/index.jsx
```diff
@@ -1,8 +1,19 @@
...
 import React from 'react'
+import { connect, Provider } from 'react-redux'
 
+import Services from '@src/Services'
 import { Dashboard } from '@src/components/Dashboard'
+import DeleteButton from '@src/components/Dashboard/DeleteButton'
 
 export const createDashboard = ({ element, ...params }) => {
   ReactDOM.render(<Dashboard {...params} />, element)
+  const deleteButton = document.body.querySelector('.export-plan-delete')
```
froddd, line 11: If there's only going to be one deletable plan on any page, would it make sense to use `id` for consistency?

react-components/src/components/Dashboard/DeleteButton.test.jsx
bobmeredith: Nice spot - No, I was fixing for the last name change and missed turning back on.

react-components/src/views/sections/Dashboard/index.jsx
bobmeredith: Good point. Will change.
Annoyingly, it seems that the designers want an EP delete from the list page too which means this whole mechanism needs to change to get any re-use.

exportplan/core/helpers.py
bobmeredith: Actually no, but then none of the endpoints do.  I created a ticket to fix this as a global decorator probably in d-api. It's with @depsiatwal for this sprint.

exportplan/templates/exportplan/dashboard_page.html
bobmeredith: Oh yes - actually, need to check with content and see if they want more detail there. Not much point in a title that matches the button label I guess.

bobmeredith: Yeah - got new titles from content 'Delete/Download your export plan'

PR #1531, opened 2021-09-06
Verify account with token

rafa-garcia:
CONTEXT: This changeset updates the account verification process to use non-PII parameters to get passed onto the query string.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3344

### Housekeeping

### Merging


core/templates/core/signup.html
```diff
@@ -13,12 +13,14 @@
 
     <script type="text/javascript">
 		var element = document.getElementById('sign-up');
-		var currentStep = '{% if request.GET.verify %}verification-code{% else %}credentials{% endif %}';
-		var email = '{{ request.GET.verify }}';
+		var uidb64 = '{{ request.GET.uidb64 }}';
+		var token = '{{ request.GET.token }}';
+		var currentStep = (uidb64 && token) ? 'verification-code' : 'cedentials';
```
stevejalim, line 18: Typo to fix?

```suggestion
		var currentStep = (uidb64 && token) ? 'verification-code' : 'credentials';
```

stevejalim: Not sure if the current wizard/workflow uses `cedentials` as its key

stevejalim (approved):
Python side LGTM (and the JS too, though that's less my thing)

core/templates/core/signup.html
bobmeredith: Looks like the wizard currently uses 'credentials' so I'm not sure how tests are passing.
`export const STEP_CREDENTIALS = 'credentials'`

stevejalim (approved):

PR #1538, opened 2021-09-09
Fix in situ code verification

rafa-garcia:
CONTEXT: This changeset incorporates a fix for a broken code verification flow when it is done in situ.

The code verification form needs a `uib64` and a `token` to submit via the `directory-sso-api-client` which are supplied in the form of query parameters on the verification link. However, when the verification form is invoked from the signup wizard (step 2), the absence of query parameters made the verification process break due to the missing info.

The remediation brings the missing information in the response on user creation, so they are available to the front end client on step 2 of the journey.

### Workflow

### Housekeeping

### Merging


sso/views.py
```diff
@@ -79,7 +79,7 @@ def post(self, request):
             form_url=self.request.path,
             verification_link=self.get_verification_link(uidb64, token),
         )
-        return Response(status=200)
+        return Response({'uidb64': uidb64, 'token': token})
```
depsiatwal, line 82: is it possible to have a test on this view to ensure code is tested ?

depsiatwal (commented):
LGTM just small comment regarding testing

depsiatwal (approved):
LGTM

PR #1547, opened 2021-09-15
Show error when End date precedes start date

michalk01:

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3346

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/core/sass/components/_forms.scss
+++ b/core/sass/components/_forms.scss
@@ -252,6 +252,13 @@
     }
   }
 
+  &__error {
+    margin-top: $gutter-small;
+    color: $great-red;
+    border-left: 5px solid $great-red;
+    padding: 20px 5px 20px 20px;
+  }
+
   & .form-group, & .select {
     margin-bottom: 0;
   }
--- a/react-components/src/components/Form/MonthYearInput/index.jsx
+++ b/react-components/src/components/Form/MonthYearInput/index.jsx
@@ -1,4 +1,4 @@
-import React, { memo } from 'react'
+import React, { memo, useState } from 'react'
 import * as PropTypes from 'prop-types'
 
 import { Input } from '../Input'
@@ -13,6 +13,8 @@ export const MonthYearInput = memo(
      yearValue,
      onChange,
      className,
+     minMonth,
+     minYear,
    }) => {
     const MONTHS = [
       'January',
@@ -29,10 +31,33 @@ export const MonthYearInput = memo(
       'December',
     ]
 
+    const [showError, setShowError] = useState(false)
+
     const monthsOptions = MONTHS.map((month, i) => ({
       label: month,
       value: `${i + 1}`,
     }))
+
+    const handleOnChange = (item) => {
+      onChange(item)
+
+      if(!minYear || !minMonth || (item[yearName] && item[yearName] < 1000)){
+        onChange(item)
+        return
+      }
+
+      const startDate = new Date(minYear, minMonth - 1);
+      const completeYear = item[yearName] ? item[yearName] : yearValue;
+      const completeMonth = item[monthName] ? item[monthName] : monthValue;
+      const completeDate = new Date(completeYear, completeMonth - 1);
+
+      if(completeDate >= startDate) {
+        setShowError(false)
+
+      } else {
+        setShowError(true)
+      }
+    }
 
     return (
       <fieldset className={className}>
@@ -43,7 +68,7 @@ export const MonthYearInput = memo(
               label='Month'
               id={monthName}
               name={monthName}
-              update={onChange}
+              update={handleOnChange}
               options={monthsOptions}
               selected={`${monthValue}`}
             />
@@ -54,13 +79,17 @@ export const MonthYearInput = memo(
               id={yearName}
               type='number'
               value={`${yearValue || ''}`}
-              onChange={onChange}
+              onChange={handleOnChange}
               size={4}
               inputMode='numeric'
               pattern='[0-9]*'
             />
           </div>
         </div>
+
+        {showError && <div className="inputgroup__error">
+          "Complete by" date cannot procede "Start objective in" date
+        </div>}
       </fieldset>
     )
   }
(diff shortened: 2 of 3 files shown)


react-components/src/components/Form/MonthYearInput/index.jsx
```diff
@@ -29,11 +31,34 @@ export const MonthYearInput = memo(
...
     const monthsOptions = MONTHS.map((month, i) => ({
       label: month,
       value: `${i + 1}`,
     }))
 
+    const handleOnChange = (item) => {
+      onChange(item)
+
+      if(!minYear || !minMonth || (item[yearName] && item[yearName] < 1000)){
+        onChange(item)
```
bobmeredith, line 45: Looks like you're calling onChange twice

react-components/src/components/Form/MonthYearInput/index.jsx
```diff
@@ -29,11 +31,34 @@ export const MonthYearInput = memo(
...
+        return
+      }
+
+      const startDate = new Date(minYear, minMonth - 1);
+      const completeYear = item[yearName] ? item[yearName] : yearValue;
+      const completeMonth = item[monthName] ? item[monthName] : monthValue;
+      const completeDate = new Date(completeYear, completeMonth - 1);
+
+      if(completeDate >= startDate) {
```
bobmeredith, line 54: Can just `setShowError(completeDate >= startDate)` rather than conditional

bobmeredith (changes requested):
A couple of things.

- Javascript/scss code changes should include built js and css.  Run 'npm run build'
- You should add a changelog entry (there is a checkbox in the PR tasklist)
- Please add a brief description of the work to the PR description, and maybe a screen grab is appropriate in this case.
- Also, the implementation. So this code will check for error when updating the end date, but not the start date. I'd suggest triggering the check on change of either as well as perhaps on initial rendering of the page.  You are preventing saving of invalid data so i guess the latter is not strictly needed.

michalk01:
error

core/sass/components/_forms.scss
```diff
@@ -252,6 +252,13 @@
     }
   }
 
+  &__error {
+    margin-top: $gutter-small;
+    color: $great-red;
+    border-left: 5px solid $great-red;
+    padding: 20px 5px 20px 20px;
```
bobmeredith, line 259: Actually we have styling for a panel with a LH bar. I would have though, this should just extend that - to override the border colour.
This is in the great-styles repo. You can see its storybook here https://great-styles-staging.london.cloudapps.digital/?path=/story/structure-text-in-a-box--with-side-highlight-bar
To go that way - you would need to update great-styles and add a story to storybook, then PR that, get it merged and then rebuild in great-cms.

react-components/src/components/Form/MonthYearInput/index.jsx
```diff
@@ -34,6 +39,29 @@ export const MonthYearInput = memo(
       value: `${i + 1}`,
     }))
 
+    useEffect(() => handleShowError({}), [])
+
+    const handleOnChange = (item) => {
+      onChange(item)
```
bobmeredith, line 45: Looks like you're calling the outer onChange in all situations.  I thought you wanted to bail on error?

bobmeredith (changes requested):
No compiled js/css or map files in here yet. Also, no description.
You should add jest unit test coverage. (npm run test). Our coverage is not perfect, but good to keep it as high as possible.
I would suggest looking hard at moving the validation check to the level above the control. I think it would be significantly simpler and less code.

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -65,7 +70,14 @@ const fwRefObjective = forwardRef((props, ref) => {
             yearValue={data.end_year}
             onChange={onChange}
             className="m-t-s"
+            minMonth={data.start_month}
+            minYear={data.start_year}
+            setShowError={setShowError}
           />
+
+          {showError && <div className="inputgroup__error">
+            "Complete by" date cannot procede "Start objective in" date
```
bobmeredith, line 79: Typo!

react-components/src/components/Form/MonthYearInput/index.jsx
michalk01: based on the discussion we I had with designer I should call it all the time

react-components/src/components/ObjectivesList/Objective/index.jsx
```diff
@@ -11,11 +11,20 @@ import { MonthYearInput } from '../../Form/MonthYearInput'
...
+
   const onChange = (item) => {
-    handleChange({
+    const updatedData = {
       ...data,
       ...item,
-    })
+    }
+
+    handleChange(updatedData)
```
bobmeredith, line 26: I don't think you want to call handleChange without checking for errors first. That does the save remember.

react-components/src/components/ObjectivesList/Objective/Objective.test.jsx
```diff
@@ -159,4 +159,34 @@ describe('Objective', () => {
 
     expect(mockFunction).toHaveBeenCalledWith(234)
   })
+
+  test('should show Error when start date precede end date', () => {
+    const dummyObjective = {
+      ...mockObjectiveData, start_year: 2021, start_month: 5, end_month: 4, end_year: 2021
+    }
+
+    console.log('dummyObjective', dummyObjective)
```
bobmeredith, line 168: console.log

react-components/src/components/Form/MonthYearInput/index.jsx
bobmeredith: Hmm, really? that means that you would save invalid data to the export plan. That seems quite illogical.

react-components/src/components/ObjectivesList/Objective/index.jsx
bobmeredith: Ah yes - deliberate but as discussed - the BE validation will blow up anyway - so you should definitely not call handleChange in an error situation.

seijihg:
have you npm run build? there should not be conflicts like these. remember to turn off npm run watch before PR. Some tests are failing.

PR #1552, opened 2021-09-22
GP2-3192 Lesson case study logic

bobmeredith:
Revamp of case study logic to allow multiple products and markets.
For performance, this uses elasicsearch to return a rough-cut of the matching case-studies along with their tags. It then applies the ranking mechanism using the weightings set through wagtail admin to determine the best CS to show.
There is a management command to load/reload the index from the db case studies. Also, the index is updated whenever a CS is added or edited.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3192

### Reviewing help

### Housekeeping

### Merging


core/blocks.py
```diff
@@ -242,21 +240,49 @@ class Meta:
...
+                        export_markets=export_markets,
+                        export_regions=export_regions,
+                        page_context=page_context,
+                        export_blocs=export_blocs,
+                        settings=settings,
+                    )
+                    hits.append(hit_dict)
+                sorted_cs_list = sorted(hits, key=lambda hit: hit.get('score'), reverse=True)
+                best_case_study = sorted_cs_list and sorted_cs_list[0]
+                if best_case_study and int(best_case_study.get('score')) >= getattr(settings, 'threshold'):
```
depsiatwal, line 280: what happens if threshold doesn't exist or not set do we have default.
Also wondering if it's worth marking more explicit  i.e ELASTIC_SEARHC_CASESTUDY_THRESHOLD

core/blocks.py
```diff
@@ -242,21 +240,49 @@ class Meta:
...
+                    best_case_study_id = best_case_study.get('pk')
+                    context['case_study'] = models.CaseStudy.objects.get(id=best_case_study_id)
+        except ConnectionError:
+            # nothing we can do without elasticsearch so continue without case study
```
depsiatwal, line 284: can see why this is here but i'm not super keen on swallowing exceptions. What happens if we got bad credentials we won't know until issue is reported. May be at very least log the error so sentry picks this up.

core/case_study_index.py
```diff
@@ -0,0 +1,71 @@
...
+        out.append(f'{page_type_mapping.get(lesson.specific._meta.model_name)}_{str(lesson.id)}')
+    return ' '.join(out)
+
+
+def stringify_tags(tags):
+    return ' '.join([str(item).replace(' ', '_') for item in tags])
+
+
+def get_connection():
+    return connections.create_connection(hosts=['localhost'])
```
depsiatwal, line 34: why local host ? is that always the case ?

depsiatwal (commented):
wow looks very good 
have very little to say :-)
by I do wonder and would have been better if we could have it in d-api 
but guess it's not possible since case studies sit in wagtail.

core/blocks.py
bobmeredith: Threshold is a admin-settable parameter rather than an environment var and it has a default - in the model. You're absolutely right though - I can access directly. A bit of copy'n'paste without thinking.

core/blocks.py
bobmeredith: OK - good plan. Yeah - this is present only so the page doesn't break if elastic-search is not available.

core/case_study_index.py
bobmeredith: That I don't know. I guess I can call create_connection without any parameters?

depsiatwal (approved):
LGTM

PR #1558, opened 2021-09-27
GP2-3257 Fix costs and pricing debounce

bobmeredith:
The current implementation of front <-> back communications is giving grief.
ATM, each character press on a field on the page, sends to the BE, only the field being edited. To prevent very large numbers of updates being sent, these requests are given a basic delay/cancel/refresh debounce with a delay of 2 seconds.
i.e. a post is delayed by 2 seconds. If another post is triggered during that time, the first post is cancelled and the timer reset for the new post.
This has the effect that:
a) while the user keeps typing, no posts are sent.
b) if the user changes multiple fields in succession, all but the last update will be lost.

To get round these issues, this PR replaces the standard debounce with a keyed version. Each post is sent with a key that identifies the field being altered. further updates to _that field_ behave as before  - cancelling previous posts. Updates to a different field though do not cancel the previous post, and so allow the update to proceed

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3257

### Reviewing help

### Housekeeping

### Merging


bobmeredith:
Testing.
Type data into a number of different fields on the C&P page of EP in rapid succession (don't stop typing for more than 2 seconds). Check that totals reflect the typed values and that everything saved OK by refreshing the page.

depsiatwal (approved):
even though not from end dev
changes look good and I trust the code :-)

PR #1666, opened 2021-11-19
GP2-3856 - Core components JavaScript build

bobmeredith:
First part of migrating all builds to webpack.
the aim is a single build to cover all Magna and logged-out sites, so ideally, we want 4 files - js and css for each site.
Sadly, the domestic and components css builds produce many separate .css files, each named from the entry scss file.
It's a big job to consolidate this, so this PR covers just the components build from core/components and a main .css build from domestic/sass.
I've renamed the distribution files to make them simpler.

**Magna build:**
unchanged apart from outputs being renamed
magna.js, magna_styles.css

**/core/components:**
components.js and components_styles.css

**/domestic/:**
loggedout.js, loggedout_styles.css

with corresponding maps 

the core/components build is actually just the cookie modal.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GP2-3856

### Reviewing help

### Housekeeping

### Merging


core/components/gulpfile.js
```diff
@@ -27,7 +27,7 @@ gulp.task(
 )
 
 // // Export-elements-specific component styling
-
+/* No longer needed
```
froddd, line 30: Get rid?

react-components/src/components/CookiesModal/CookiesModal.test.jsx
```diff
@@ -25,6 +23,9 @@ const defaultProps = {
 describe('CookiesModal', () => {
 
   test('handles not being shown', () => {
+    global.document.cookie = "cookie_preferences_set=true"; //set cookies in whatever format you like
+    document.cookie = "cookie_preferences_set=true";
+    CookiesManager.getPreferencesCookie.mockImplementation(() => true)
```
froddd, line 28: If `getPreferencesCookies` is mocked, is it necessary to actually place a cookie in the test `document`? Seems like one of these only would signify intent a bit better.

react-components/src/components/CookiesModal/dit.components.cookie-notice.js
```diff
@@ -120,6 +120,7 @@ dit.components.cookieNotice = function() {
   }
 
   function getPreferencesCookie () {
+    console.log('Cookies - in gerPreferencescookie', document.cookie, getCookie(cookiePreferencesName))
```
froddd, line 123: Probably remove :)

react-components/webpack.config.js
```diff
@@ -49,9 +53,10 @@ module.exports = {
         exclude: /node_modules/,
       },
       {
-        test: /\.s?css$/i,
+        test: /\.scss$/i,
+        exclude: /node_modules/,
         use: [
-          // extract to seperate file
+          // extract to separate file
```
froddd, line 59: 🙌 😄

seijihg:
there is a licence file added.

froddd (commented):
A couple of minor code cleanup tweaks.

Not sure if related, but the cookie preferences page doesn't quite seem to work as intended: when loading after accepting all cookies, none of the preferences are shown. Selecting any combination of preferences seems to always produce (or retain?) the same preferences in the cookie:

```
{"essential":true,"settings":true,"usage":true,"campaigns":true}
```

react-components/webpack.config.js
bobmeredith: You'll always find a rat in separate!

react-components/src/components/CookiesModal/CookiesModal.test.jsx
bobmeredith: I remember struggling with this - there seemed to be some cross-contamination between tests.
I'll have a look.

froddd (approved):
Code looks good -- still an issue with cookie preferences.

core/components/gulpfile.js
bobmeredith: Yeah - I was always intending to kill the whole file

bobmeredith:
@froddd  Yeah, thanks for checking that. There was an issue - I needed to include an extra file in the components bundle.
All better now.

PR #1727, opened 2022-02-03
Ensure focus styles exist

froddd:
Make sure carousel buttons render correctly and are focusable. Previous removal of unnecessary theme had applied dubious defaults!


webbyfox (approved):
LGTM

PR #1822, opened 2022-08-04
Redirect unverified user to verification code page after login

mforner13:
### Description
This PR makes an update to the login flow to make it easier for users who have previously signed up, but not yet verified their account, to login.

**Previous flow:**
- Unverified (exists in the db but hasn't entered a verification code) user enters their details into the login form
- They are redirected to `/dashboard`
- Because the dashboard view is protected by the `AuthenticatedUserRequired` mixin, they are then redirected back to `/signup`

**New flow:**
- Unverified user enters their details into the login form
- Directory SSO returns a 401 unauthorised response
- This status code triggers the re-issuing of a verification code email with a new code attached
- The `uidb64` and `verification_token` values are taken from the new code which is created, and returned in the response object
- The `handleSubmit` function in the client code determines the url to redirect to based on the existence of these response parameters
- The user is then redirected to the same URL which would be in the verification code email, e.g. `http://greatcms.trade.great:8020/signup/?uidb64=OA&token=636-5298c48e1e68b7946672`

### To test locally
- Run against the `feature/unverified-user-login-flow` branch in `directory-sso`
- Create a new user via `/signup`
- Do not verify that user
- Go to `/login` and enter the details for the new user
- Check that you are redirected to the verification code step of the signup flow

### Screenshots

Screenshot 2022-08-04 at 11 58 55

### Workflow

### Reviewing help

### Housekeeping

### Merging


react-components/tests/views/LoginModal/Wizard.test.jsx
```diff
@@ -20,25 +19,15 @@ describe('LoginModal', () => {
...
     window.location = { assign: jest.fn() }
-
-    Services.setConfig({
-      apiLoginUrl: 'http://www.example.com/login/',
-      csrfToken: '123',
-      linkedInUrl: 'http://www.example.com/oauth2/linkedin',
-      googleUrl: 'http://www.example.com/oauth2/google',
-      dashboardUrl: '/dashboard/',
-    })
   })
```
mforner13, line 22: This function call wasn't previously doing anything, as the whole Services module was being mocked at the top of the file. Only the signupUrl is needed from config, so this is now set in the individual tests that need it.

sso/helpers.py
```diff
@@ -53,7 +53,7 @@ def get_cookie(cookie_jar, name):
 
 def response_factory(upstream_response):
     cookie_jar = get_cookie_jar(upstream_response)
-    response = Response(status=200)
+    response = Response({}, status=200)
```
mforner13, line 56: Now that `response.json()` is called on the response returned from `SSOBusinessUserLogin`, the data property of the response can't be empty, as the promise can't resolve when it is empty.

tests/unit/sso/test_views.py
```diff
@@ -49,6 +49,22 @@ def test_business_sso_login_500_upstream(client, requests_mock):
...
+def test_business_sso_login_401_upstream(mock_send_code, mock_regenerate_code, client, requests_mock):
+    mock_regenerate_code.return_value = {'code': '12345', 'user_uidb64': 'aBcDe', 'verification_token': '1ab-123abc'}
+    requests_mock.post(settings.SSO_PROXY_LOGIN_URL, status_code=401)
+
+    response = client.post(reverse('sso:business-sso-login-api'), {'email': 'test', 'password': 'password'})
+
+    assert mock_send_code.call_count == 1
+    assert mock_regenerate_code.call_count == 1
+
+    assert response.status_code == 200
```
webbyfox, line 64: very confused with this test method, Name suggests you are expecting 401 but asserting 200.

sso/views.py
```diff
@@ -24,6 +40,20 @@ def post(self, request):
             'login': serializer.validated_data['email'],
         }
         upstream_response = requests.post(url=settings.SSO_PROXY_LOGIN_URL, data=data, allow_redirects=False)
+        # 401 means credentials are correct, but user is unverified
+        if upstream_response.status_code == 401:
```
webbyfox, line 44: I would personally use 403 or 422 here

webbyfox (commented):
Overall looks good, just minor comments

tomstevens9 (approved):

tests/unit/sso/test_views.py
mforner13: The test name is talking about handling a 401 **upstream** response (from directory sso). However if a user's credentials are correct, the login view in this repo still returns a 200 response, but also generates a new verification code.

sso/views.py
mforner13: I went with 401 as it indicates that they are authenticated but unauthorised to access the resource. 403 didn't seem right as the user is authenticated as they've provided correct credentials, and then 422 seemed to vague. What do you reckon?

webbyfox: I think 403 is good

webbyfox (approved):

PR #1843, opened 2022-08-18
Update signup to resend verification code is current one has expired

tomstevens9:
This commit modifies the sign up flow so that if an expired verification code is entered by the user then a new code is sent to them. They are also presented with a message in the UI to inform them that they've been sent a new code.

To test
* Create a new user account in great-cms
* Wait 30 minutes for verification code to expire*
* Enter expired verification code
* Check to see if new verification code has been sent
* Enter new verification code

\* If testing in local environment then sso/verification/models.py in directory-sso can be modified to reduce the code expiry time.

Before and after for UI changes:
Screenshot 2022-08-19 at 09 48 01
Screenshot 2022-08-19 at 09 34 04

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GLS-336

### Reviewing help

### Housekeeping

### Merging


react-components/src/components/Form/FormGroup/index.jsx
```diff
@@ -20,13 +20,15 @@ export const FormGroup = memo(
...
     info,
+    message,
   }) => {
     return (
       <div
         className={`form-group ${
           errors.length > 0 ? 'form-group-error' : ''
         } ${formGroupClassName}`}
       >
+        {message ? <div className="g-panel body-m m-v-xs">{message}</div> : null}
```
mforner13, line 31: What was the reason for adding the new message prop rather than using the errors prop? :)

mforner13 (commented):
Looks good! Just one question

react-components/src/components/Form/FormGroup/index.jsx
tomstevens9: There's a short discussion on the ticket about this. In short, the feeling was that this message isn't an error and might give the user the wrong impression that something is wrong.

mforner13 (approved):

PR #1938, opened 2022-11-25
Correct contact CTA destinations for Scottish trade offices

mforner13:
This PR fixes an error in which the wrong destinations were applied to the Scottish enterprise and DIT Scotland office's contact buttons.

Since both offices need to have a region of `scotland` in directory API to be matched with Scottish postcodes, the logic now checks the name of the office, and if it is the Scottish Enterprise office, uses their website URL as the destination of the contact button.

The styling of the Scottish Enterprise contact us link has also been made consistent with all other offices - showing a button rather than a link. Shown below.

**TODO**
If more offices ask to be contacted via their website rather than our contact form, we should add a field to the [trade office model in directory-api](https://github.com/uktrade/directory-api/blob/develop/exporting/models.py#L6) which sets whether the office wants to be contacted via their website or our form. The template in this repo would then use that logic to determine the destination of the contact button, rather than it being hardcoded to the office name.

**Screenshot - after**
Screenshot 2022-11-25 at 15 37 35

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-80

### Reviewing help

### Merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,17 @@
 ## Pre release
 
 ### Bugs fixed
+* KLS-80 - Correct form destinations and CTA style for Scottish offices
 
 ### Hotfix
 
 ### Enhancements
+
+## [2.19.0](https://github.com/uktrade/great-cms/releases/tag/2.19.0)
+### Enhancements
 * KLS-124 - Remove FTA feature flag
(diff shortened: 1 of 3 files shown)


hal274 (approved):
LGTM

PR #1952, opened 2022-12-09
Kls 194 upgrade django wagtail

bsabhilash:
Upgrade django to 3.2
wagtail to 2.15


tests/unit/core/test_blocks.py
```diff
@@ -53,12 +53,6 @@ def test_modular_content_static_block_render():
     assert html == expected_html
 
 
-def test_render_form_with_constructor():
```
bsabhilash, line 56: Test removed. render_form is not available in wagtail 2.13 onwards.
See this https://stackoverflow.com/questions/68080777/how-is-a-fieldblock-rendered-in-wagtail-cms-2-13-x

tests/unit/activitystream/test_views.py
```diff
@@ -358,7 +358,7 @@ def test_pagination(
     # TODO: Improve performance of page.url, full_url, full_path
     # Since page.url needs to get the slugs of the article's parent
     # pages it is doing a TON of queries each time this endpoint is hit
-    with django_assert_num_queries(67):
+    with django_assert_num_queries(64):
```
mforner13, line 361: Do you know why this is now reduced?

bsabhilash: Not completely sure what happens behind the scenes so that count comes to 67 or 64 now.

I think that is not relevant either way. The test is to check if pagination works, it is working as confirmed by other asserts, during the pagination there are django_calls which are counted here. 
50 calls for 50 articles and few more. not sure how it comes out to 64 now and not 67.

PR #1975, opened 2023-01-31
Feature/flexible related pages fields

ecbm:
This PR adds two additional fields to each of the related pages in wagtail article pages. The Title and URL fields allow you to add an external link to the related pages area of any article page. Setting these variables will override any currently selected pages.

With this change we are also increasing the number of allowed related pages from 3 to 5.

Screenshot 2023-01-31 at 15 29 10

If you encounter a migration error when running tests locally: 
switch over to the test database which should be called test_[your database name], run the migrations then return the original database name and rerun tests.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/212

### Reviewing help

### Merging


domestic/cms_panels.py
```diff
@@ -133,13 +133,42 @@ class ArticlePagePanels:
             ],
         ),
         MultiFieldPanel(
-            heading='Related articles',
+            heading='Related pages',
+            help_text='Entering a title and link overrides the page choice',
             children=[
+                PageChooserPanel('related_page_one', 'wagtailcore.Page'),
```
bsabhilash, line 139: Does this have to be 5 hardcoded list. Can we populate a dynamic list so content owners can add as many more or less they need?
Examples are found through out the app.

ecbm: For this ticket there was a hard limit on the number of items a content designer should be able to add. So in this ticket I raised it from 3 to 5 chooser panels max. Although it would probably be cleaner to have a dynamic list with a limit of 5! But I wasn't too confident changing the pattern from what was there before.

bsabhilash: If you want to try out this locally,

wrap first child to `MultiFieldPanel` e.g. :https://github.com/uktrade/great-cms/blob/37e152e2bd9762839dabda407507dae4ca81b601/domestic/cms_panels.py#L200

and

set max/min limit in model e.g.: https://github.com/uktrade/great-cms/blob/37e152e2bd9762839dabda407507dae4ca81b601/domestic/models.py#L286

ecbm: Would this effect how the related page columns are named/populated? I worry about how this would effect the related pages that are already in the db, I need to make sure the new changes are compatible with the fields that are already present

hal274: I like the dynamic list with a limit of 5

bsabhilash (approved):

PR #1978, opened 2023-02-02
Feature/ioo setup

stuart-madetech:
Initial setup of new International Online Offer project within great-cms. Routing changed to support new project, with branding from international UI.  

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IOO-399

### Reviewing help

Before:
N/A
After:
Screenshot 2023-02-02 at 12 25 20

### Housekeeping

### Merging


core/context_processors.py
```diff
@@ -67,3 +70,28 @@ def feature_flags(request):
     retval = {'features': {}}
     retval['features'].update(settings.SSO_PROFILE_FEATURE_FLAGS)
     return retval
+
+
+def footer_contact_us_link(request):
+    if settings.FEATURE_FLAGS.get('INTERNATIONAL_CONTACT_LINK_ON'):
```
mforner13, line 76: Just checked how this feature flag is used in great-international and looks like it was added 3 years ago, and is set to True in great-international-ui production.
I think from this we're safe to remove the if statement and just use `footer_contact_us_link = urls.international.HOME / 'contact'`

core/context_processors.py
```diff
@@ -67,3 +70,28 @@ def feature_flags(request):
...
+        footer_contact_us_link = urls.international.HOME / 'contact'
+    else:
+        footer_contact_us_link = urls.international.CONTACT
+
+    return {'footer_contact_us_link': footer_contact_us_link}
+
+
+def directory_components_html_lang_attribute(request):
+    return {'directory_components_html_lang_attribute': translation.get_language()}
```
mforner13, line 85: This is used in the directory-components base.html which the ioo template extends from, so looks like we should keep it in. However we should also bring the corresponding test for it over from great-international-ui -[test here](https://github.com/uktrade/great-international-ui/blob/develop/core/tests/test_context_processors.py#L32).

core/context_processors.py
mforner13: Scrap that - I've just looked and this context variable `footer_contact_us_link` is actually no longer used in the footer template in great-international-ui! Found where it was used in the [initial commit](https://github.com/uktrade/great-international-ui/commit/268dbea6aaac4a1b45b798a777a1aa39961a8bca) for the repo, but has now been replaced with `services_urls.contact_us_international` which is also used in the ioo footer. 

So you can remove this whole block :)

core/context_processors.py
```diff
@@ -67,3 +70,28 @@ def feature_flags(request):
...
+
+
+def services_home_links(request):
+    return {
+        'international_home_link': {'url': reverse_lazy('index'), 'label': _('great.gov.uk international')},
+        'trade_home_link': {'url': reverse_lazy('find-a-supplier:trade-home'), 'label': _('Find a supplier')},
+        'investment_atlas_home_link': {'url': reverse_lazy('atlas-home'), 'label': _('Invest in the UK')},
+    }
```
mforner13, line 93: These two context variables shouldn't be needed, they relate to the wider international site, but won't be linked to from IOO. If they are needed in the future we can always add them in :)

`international_home_link` **is** used in the the ioo base.html though :)

core/context_processors.py
```diff
@@ -67,3 +70,28 @@ def feature_flags(request):
...
+
+
+def header_navigation(request):
+    return {'navigation_tree': header_config.nav_tree.ATLAS_HEADER_TREE}
```
mforner13, line 97: Is the atlas header tree going to be used in IOO?

mforner13 (commented):
Haven't gone through everything yet, but some changes to start with :)

core/context_processors.py
stuart-madetech: Removed block

core/context_processors.py
stuart-madetech: Removed the two mentioned.

core/context_processors.py
stuart-madetech: Removed as not using just yet

core/context_processors.py
stuart-madetech: Added test and passes

international_online_offer/static/js/dit.cookieBanner.js
```diff
@@ -0,0 +1,76 @@
+dit = window.dit || {}
```
mforner13: It looks like this isn't used in any templates so I think you can remove :) (please correct me if I'm wrong!)

international_online_offer/urls.py
```diff
@@ -0,0 +1,23 @@
...
+
+from international_online_offer import views
+
+SIGNUP_URL = reverse_lazy('core:signup')
+# NB our signup/signin redirection workflow following login_required
+# relies on the value of REDIRECT_FIELD_NAME being the default: 'next'
+# If you change the redirection parameter, other code will need
+# updating too such as core.wagtail_hooks.authenticated_user_required,
+# core.templatetags.url_tags.get_intended_destination and the loginUrl
+# and signupUrl in base.html
```
mforner13, line 11: You're not referencing `SIGNUP_URL` in this file so can remove this section :)

international_online_offer/urls.py
```diff
@@ -0,0 +1,23 @@
...
+
+app_name = 'international_online_offer'
+
+urlpatterns = [
+    # Temp redirect to old dashboard this can be removed over time this is to allow bookmarks and other services
+    # To change the base dashboard link which is partially controlled by directory-constants
```
mforner13, line 17: I think can remove this comment as is taken from another urls file

international_online_offer/templates/ioo/base.html
```diff
@@ -0,0 +1,58 @@
+{% extends 'directory_components/base.html' %}
```
mforner13, line 1: Because all the templates in `international_online_offer/templates/` will related to ioo, I don't think you need the extra `ioo/` folder :) Can just be `international_online_offer/templates/base.html`. If this app grows a lot in the future we can always split the templates folder up then.

mforner13 (commented):
Last comments :)

international_online_offer/templates/ioo/base.html
stuart-madetech: This is a weird one. If I remove the IOO folder under templates the UI header and footer defaults to domestic

mforner13 (approved):
Looks good - first PR woohoo 🎉

PR #1989, opened 2023-02-09
Export Academy app

bsabhilash:
This changeset includes a bare-bones, first iteration of the Export Academy app.

The app is behind a feature flag. To enable it, the `FEATURE_EXPORT_ACADEMY` env var needs to be set to true in the application's environment like this `FEATURE_EXPORT_ACADEMY=true`.

image
image
image
image
image

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-268

### Reviewing help

- [X] Includes screenshot(s)

### Housekeeping

- [X] Added all new environment variables to Vault.

### Merging


export_academy/forms.py
```diff
@@ -0,0 +1,98 @@
...
+    job_title = forms.CharField(
+        label=_('Job title'),
+        max_length=50,
+        error_messages={
+            'required': _('Enter your job title'),
+        },
+    )
+    business_name = forms.CharField(
+        label=_('Business name'),
+        max_length=50,
```
mforner13, line 39: I think 50 could be too little for a business name - is there a reason to limit it this low?

export_academy/forms.py
```diff
@@ -0,0 +1,98 @@
...
+        choices=COUNTRIES,
+    )
+    like_to_discuss = forms.ChoiceField(
+        label=_('Radio button example'),
+        choices=(
+            ('no', 'No'),
+            ('yes', 'Yes'),
+        ),
+        widget=forms.RadioSelect,
+        error_messages={'required': _('Please answer this question')},
```
mforner13, line 69: This error message needs to follow the standard gov uk design - here's the guidance for radios - https://design-system.service.gov.uk/components/radios/#error-messages.

Might be best to double check with Neil :)

export_academy/models.py
```diff
@@ -0,0 +1,99 @@
...
+        MultiFieldPanel(
+            heading='Details',
+            children=[
+                FieldPanel('name'),
+                FieldPanel('description'),
+                FieldPanel('link'),
+            ],
+        ),
+        MultiFieldPanel(
+            heading='Date',
```
mforner13, line 40: Could `'Timing'` or `'Dates'` be a better heading here as there are multiple dates?

export_academy/models.py
```diff
@@ -0,0 +1,99 @@
...
+    CONFIRMED = 'Confirmed'
+    CANCELLED = 'Cancelled'
+    STATUSES = (
+        (CONFIRMED, CONFIRMED),
+        (CANCELLED, CANCELLED),
+    )
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    event = ParentalKey(Event, on_delete=models.CASCADE, related_name='bookings')
+    registration = models.ForeignKey(Registration, on_delete=models.CASCADE)
```
mforner13, line 87: Does this mean the person who has signed up for the event? If so I think `attendee` could be a clearer field name

export_academy/models.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    class Meta:
+        ordering = ('-created', '-modified')
+
+    def __str__(self):
+        return f'{self.id}:{self.name}'
+
+
+class Registration(TimeStampedModel):
```
mforner13, line 64: I think this name implies it brings together one user registering for one event - which is what the Booking model does. What about `EventsUser`, `EventUser` or `Attendee` ?

export_academy/forms.py
bsabhilash: We don't know yet. Copied from the existing company registration form. This is a draft, and we don't even know if this field is needed. I will clarify when the requirements are confirmed.

export_academy/forms.py
bsabhilash: Thanks for the pointer. I'll make sure to clarify with Neil. We do not know whether a radio select is needed at this stage.

mforner13 (approved):

PR #2003, opened 2023-02-17
Export Academy booking/cancellation notification

bsabhilash:
Gov notify for booking and cancellations

### Workflow

### Merging


export_academy/urls.py
```diff
@@ -32,7 +32,7 @@ def registration_required(function):
         name='registration-success',
     ),
     path(
-        'booking/success/',
+        'booking/success',
```
mforner13, line 35: Not sure if this matters but all the other URLS end with a `/` - should we keep that pattern? :)

mforner13 (approved):
LGTM 👍 Tiny comment, but I won't block :)

rafa-garcia (approved):
LGTM

export_academy/urls.py
bsabhilash: Ah, good point. I wonder why I changed this. Either way, the link will change in future when we tackle the booking confirmation page. This ticket was all about booking notification

PR #2013, opened 2023-02-24
Feature/ioo 415 triage question 2

stuart-madetech:
This is the second page in the triage for international online offer

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IOO-415

### Reviewing help

Before:
N/A
After:
Screenshot 2023-02-24 at 15 45 49

### Housekeeping

### Merging


international_online_offer/forms.py
```diff
@@ -15,3 +15,21 @@ class SectorForm(forms.Form):
...
+        ('Find people with specialist skills', 'Find people with specialist skills'),
+        ('Other', 'Other'),
+    ]
+    intent = forms.fields.MultipleChoiceField(
+        label='',
+        required=True,
+        widget=forms.CheckboxSelectInlineLabelMultiple(attrs={'id': 'intent-select'}),
+        choices=CHOICES,
+    )
+    intent_other = forms.CharField(label='', min_length=2, max_length=50, required=False)
```
mforner13, line 35: If this field is required when a user has selected the 'Other' option, I think you need to add some validation like in [this example](https://github.com/uktrade/directory-api/blob/develop/survey/models.py#L99). The example overrides the `clean_fields` method in a model, but for a form you would override the `clean()` method of the form class :)

[Useful django docs on this](https://docs.djangoproject.com/en/4.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other)

international_online_offer/templates/ioo/triage/intent.html
```diff
@@ -0,0 +1,138 @@
...
+          companyTypeOther
+      );
+  </script>
+  <script type="text/javascript">
+    const checkboxContainer = document.getElementById('intent-select');
+    const checkboxElements = checkboxContainer.querySelectorAll('input[type=checkbox]');
+    const otherCheckboxElement = document.getElementById('intent-select_5');
+    otherCheckboxElement.addEventListener("change", (e) => {
+      if (otherCheckboxElement.checked == false) {
+        document.getElementById('id_intentOther').value = '';
```
mforner13, line 108: I think this should be `id_intent_other` - it looks like that's what the id of the 'other' text field comes up as in the browser . And the same on lines 118 and 132
Screenshot 2023-02-27 at 14 34 08

international_online_offer/templates/ioo/triage/intent.html
```diff
@@ -0,0 +1,138 @@
...
+
+      for (let i = 0; i < previouslySelectedIntent.length; i++) {
+        for (let o = 0; o < checkboxElements.length; o++) {
+          if (previouslySelectedIntent[i] == checkboxElements[o].value) {
+            checkboxElements[o].checked = true;
+          }
+        }
+      }
+      if (window.localStorage.getItem('intentOther') != null) {
+         document.getElementById('id_intentOther').value = window.localStorage.getItem('intentOther');
```
mforner13, line 132: This gets it out of local storage, but doesn't look like it makes the text input field appear when a user comes back to this page.
Screenshot 2023-02-27 at 14 37 04

international_online_offer/templates/ioo/triage/intent.html
```diff
@@ -0,0 +1,138 @@
...
+            </div>
+        </div>
+      </main>
+    </div>
+{% endblock %}
+
+{% block body_js %}
+  {{ block.super }}
+  <script type="text/javascript" src="{% static 'javascript/dit.components.toggleOther.js' %}"></script>
+  <script type="text/javascript">
```
mforner13, line 91: I think as a general rule we don't want a lot of javascript in the template as it's more difficult to read and makes the files quite long. What do you think to separating this out into a separate js file and just calling the functions that we need to from the html file?

international_online_offer/templates/ioo/triage/intent.html
```diff
@@ -0,0 +1,138 @@
...
+                  {% include 'core/includes/form_field.html' with field=form.intent_other %}
+                </div>
+                <br/>
+                <button type="submit" class="atlas-button">Continue
+                  <span class="atlas-button__icon">
+                    <svg viewBox="0 0 13 20" xmlns="http://www.w3.org/2000/svg" preserveAspectRatio="xMidYMid meet" xmlns:xlink="http://www.w3.org/1999/xlink">
+                      <path fill="currentColor" d="M2.893 20l9.834-10L2.893 0 0 2.921 6.976 10 0 17.079z"></path>
+                    </svg>
+                  </span>
+                </button>
```
mforner13, line 53: I think you can extract this into an includes as it's duplicated in the sector file and I think will be used a couple more times :)

mforner13: Or even make a template for the triage steps where you can just pass in your different blocks which differ for each page - e.g. the fields and heading and additional text - let me know if you want to go through this :)

international_online_offer/forms.py
stuart-madetech: I have done this for the next step, I remember a conversation around not wanting it to be required as some users select other and dont type anything in. Will add it in anyway and can pull it out if not required

mforner13: Ahh understood! :) Best to check with your designer & PM then I'd say

international_online_offer/templates/ioo/triage/intent.html
stuart-madetech: Agreed to do this on next PR

international_online_offer/templates/ioo/triage/intent.html
stuart-madetech: Agreed to do this on next PR

international_online_offer/templates/ioo/triage/intent.html
stuart-madetech: Agreed to do this in next PR :)

international_online_offer/forms.py
```diff
@@ -15,3 +15,33 @@ class SectorForm(forms.Form):
...
+
+    def clean(self):
+        data = self.cleaned_data
+        intent = data.get('intent')
+        intent_other = data.get('intent_other')
+        if not intent:
+            self.add_error('intent', 'This field is required')
+            return data
+        if any('Other' in s for s in intent) and not intent_other:
```
mforner13, line 44: You can simplify this with just:
```
if 'Other' in intent and not intent_other:
```

international_online_offer/forms.py
```diff
@@ -15,3 +15,33 @@ class SectorForm(forms.Form):
...
+    intent = forms.fields.MultipleChoiceField(
+        label='',
+        required=False,
+        widget=forms.CheckboxSelectInlineLabelMultiple(attrs={'id': 'intent-select'}),
+        choices=CHOICES,
+    )
+    intent_other = forms.CharField(label='', min_length=2, max_length=50, required=False)
+
+    def clean(self):
+        data = self.cleaned_data
```
mforner13, line 38: Good to make sure you call the parent class' clean method first, `cleaned_data = super().clean()` 

See this [example in django docs](https://docs.djangoproject.com/en/4.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other)

international_online_offer/forms.py
mforner13: Don't need to check for `intent` first as calling `super().clean()` will check your required fields first :)

stuart-madetech: Done this in a different way - errors that intent is not iterable when null

international_online_offer/forms.py
stuart-madetech: Added this :)

PR #2042, opened 2023-03-15
Two column blocks

ecbm:
This PR adds two enables two column blocks for all stream builders where previously only 3 column blocks were possible.

On web:
Screenshot 2023-03-15 at 08 54 04

On mobile:
Screenshot 2023-03-15 at 09 06 00

This change can be tested in the ArticlePage page type and the MicrositePage page type in the wagtail cms.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/364

### Reviewing help

### Merging


core/templates/core/includes/_columns.html
```diff
@@ -16,10 +16,10 @@ <h3><a class="card-url" href="{{ item.value.link }}">{{ item.value.title }}</a><
 
     </div>
 </div>
-{% else %}
+{% elif value|length == 2 %}
 <div class="column-half-m column-full">
     <div class="card only-title no-padding-card transparent-card">
-    <div class="card-image">{% image item.value.image width-200 %}</div>
+    <div class="card-image">{% image item.value.image width-300 %}</div>
```
bsabhilash, line 22: How does this look on a mobile device? Could you please add a picture?

ecbm: Added a screenshot to the PR, The image is larger with 2 columns as apposed to 3 on mobile. Do you think it would be best to leave the image widths the same?

bsabhilash: It will be good to make them (3 column vs 2 column) similar on a mobile view.

ecbm: I'll change it so the images fill the available space in their column rather than having it be hardcoded

ecbm: I think I will need to talk to a designer about changing how the images are sized. I don't want to make a change that will effect how existing articles are displayed without knowing the intentions behind the current sizing. For now I've reverted the size change on this ticket to keep things inline with the current design.

PR #2044, opened 2023-03-15
EA Event details page

bsabhilash:
Event details page - video playback with subtitles


config/settings.py
```diff
@@ -859,9 +862,11 @@
     'MAINTENANCE_MODE_ON': env.bool('FEATURE_MAINTENANCE_MODE_ENABLED', False),  # used by directory-components
     'ADMIN_REQUESTS_ON': env.bool('FEATURE_ADMIN_REQUESTS_ENABLED', False),
 }
-# parity with nginx config for maximum request body
-DATA_UPLOAD_MAX_MEMORY_SIZE = 6 * 1024 * 1024
-FILE_UPLOAD_MAX_MEMORY_SIZE = 6 * 1024 * 1024
+# Enable large file uploads
+WAGTAILIMAGES_MAX_UPLOAD_SIZE = 500 * 1024 * 1024
```
timothyPatterson, line 866: Is this also for the video upload?

export_academy/templates/export_academy/event_list.html
```diff
@@ -48,8 +48,19 @@ <h6>Duration: {{ event.start_date|timesince:event.end_date }}</h6>
...
+                                            {% else %}
+                                                <a>There are no reference materials from this event.</a>
+                                            {% endif %}
+                                            {% if event.video_recording %}
+                                                <a href={% url 'export_academy:event-details' pk=event.pk%}>View video</a>
+                                            {% endif %}
+                                        {% else %}
+                                            <input type="hidden" name="status" value="Cancelled">
+                                            <button type="submit" class="link">Cancel</button>
+                                        {% endif %}
```
timothyPatterson, line 63: I think there is some crossover with https://github.com/uktrade/great-cms/pull/2046 - we can catch up.

export_academy/views.py
```diff
@@ -9,7 +15,7 @@
 
 class EventListView(FilterView, ListView):
     model = models.Event
-    queryset = model.upcoming
+    queryset = model.objects
```
timothyPatterson, line 18: Does this ignore the filters?

config/settings.py
bsabhilash: Not really. But will help to add big images. Added as a step to support big files.

export_academy/templates/export_academy/event_list.html
bsabhilash: Sure

export_academy/views.py
bsabhilash: Yes. Needed to show old events until Rafa's ticket to show old events are completed.

config/settings.py
timothyPatterson: OK because I don't think 500Mb is enough for videos - a one hour Teams recording is circa 400Mb.

export_academy/views.py
timothyPatterson: Should it be switched back before merging?

config/settings.py
bsabhilash: Agree; I do not fully understand this feature. Large file uploads are failing in dev. Once we have this flag in dev, we can adjust the values accordingly

timothyPatterson (approved):

PR #2046, opened 2023-03-17
EAM-104 Webinar buttons

timothyPatterson:
This PR adds (unstyled) webinar join and view buttons in addition to modifying the existing book and cancel form buttons. Key button attributes are generated programatically and subsequently rendered via an include template. The result of the EventButtonHelper class is an array of button specifications meaning that multiple buttons can be returned.

### Workflow

- [x]  https://uktrade.atlassian.net/jira/software/projects/KLS/boards/359?selectedIssue=KLS-360

### Merging


export_academy/models.py
```diff
@@ -58,6 +59,10 @@ class Event(TimeStampedModel, ClusterableModel, EventPanel):
 
     FORMAT_CHOICES = [(ONLINE, 'Online'), (IN_PERSON, 'In-person')]
 
+    NOT_STARTED = 'not_started'
```
bsabhilash, line 62: Since these are constants, should we define them elsewhere?

export_academy/views.py
```diff
@@ -13,6 +14,10 @@ class EventListView(FilterView, ListView):
     filterset_class = filters.EventFilter
     template_name = 'export_academy/event_list.html'
 
+    def get_buttons_for_obj(self, obj):
+        user = self.request.user
+        return EventButtonHelper().get_buttons_for_obj(user, obj)
```
bsabhilash, line 19: If this is only used once, should this be a function than a class method?

export_academy/helpers.py
```diff
@@ -6,7 +6,50 @@
 from django.urls import reverse_lazy
 
 from core.urls import SIGNUP_URL
-from export_academy.models import Registration
+from export_academy.models import Event, Registration
+
+
+class EventButtonHelper:
+    def get_buttons_for_obj(self, user, obj):
```
bsabhilash, line 13: Nice one. Good base for expanding its capability!

export_academy/models.py
timothyPatterson: Conceptually they are quite tightly coupled to an event instance so I thought it made sense to have here as opposed to the reader having to skip to another file to see the value. However, happy to move - where do you think?

export_academy/views.py
timothyPatterson: Sure, no probs.

export_academy/helpers.py
timothyPatterson: Thanks 😄

export_academy/models.py
bsabhilash: Agree. I am not really a fan of putting in far-off places. 2 options I can think of is
1. `from core.constants` import this the theme I can see. Either put it in code or create a EA constant file
2. Place it on top of the file, below imports.
I would say, adding a constant as part of the class is inappropriate, as it is difficult to use for anyone outside the class. but again what if no one wants to use it :)

hal274: I like the idea of keeping a few constants at top of file. However if it's begins to grow it will make the code difficult to read so at that point extracting to a separate file will be better. 

Generally, separating constants from objects is cleaner regardless of number of re-use

bsabhilash: Ok, As I am voted out, will resolve this conv.

rafa-garcia (approved):
Nice one, LGTM 👍🏻

export_academy/views.py
timothyPatterson: Changed to function.

export_academy/helpers.py
```diff
@@ -5,10 +5,74 @@
...
+                        'classname': 'link',
+                        'value': 'Cancelled',
+                        'type': 'submit',
+                    },
+                ]
+            elif event.status is EXPORT_ACADEMY_EVENT_IN_PROGRESS:
+                result['event_action_buttons'] += [
+                    {'url': event.link, 'label': 'Join', 'classname': 'text', 'title': 'Join'},
+                ]
+            elif event.completed:
```
bsabhilash, line 33: When/If the event is marked as completed while EXPORT_ACADEMY_EVENT_IN_PROGRESS, control will not reach here

export_academy/helpers.py
```diff
@@ -5,10 +5,74 @@
...
+)
 from core.urls import SIGNUP_URL
 from export_academy.models import Registration
 
 
+def get_buttons_for_event(user, event):
+    result = dict(form_event_booking_buttons=[], event_action_buttons=[])
+    if is_export_academy_registered(user):
+        if event.bookings.filter(registration_id=user.email, status='Confirmed').exists():
+            if event.status is EXPORT_ACADEMY_EVENT_NOT_STARTED:
```
bsabhilash, line 20: Control will fall through if event is not marked as complete  and time has expired. 
I am suggesting to add another if to check in the end to see event time hs expired and show no buttons (or check with PM to confirm)

core/constants.py
```diff
@@ -74,3 +74,7 @@
 USER_DATA_NAMES = {'ComparisonMarkets': 16384, 'UserProducts': 16384, 'UserMarkets': 16384, 'ActiveProduct': 256}
 
 COUNTRY_FACTSHEET_CTA_TITLE = 'View latest trade statistics'
+
+EXPORT_ACADEMY_EVENT_NOT_STARTED = 'not_started'
```
rafa-garcia, line 78: I do not think there is a case for polluting the constants in the core app just to assign values to event objects. This is going against the regular practice of defining constants in the class when only the class will make use of the values.

export_academy/helpers.py
timothyPatterson: Good catch, thanks. I have moved the if event.completed to first in control flow.

export_academy/helpers.py
timothyPatterson: No buttons is the default result so I think this will be ok.

rafa-garcia (changes requested):

export_academy/models.py
```diff
@@ -58,6 +59,10 @@ class Event(TimeStampedModel, ClusterableModel, EventPanel):
 
     FORMAT_CHOICES = [(ONLINE, 'Online'), (IN_PERSON, 'In-person')]
 
+    EVENT_NOT_STARTED = 'not_started'
```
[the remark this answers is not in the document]
timothyPatterson: No worries @rafa-garcia , I think this is a good comment and have implemented.

rafa-garcia (approved):

PR #2049, opened 2023-03-20
KLS-279: campaign homepage slice

lewis-coulson-dit:
Creates a new 3 card section on homepage just below hero banner.

### Workflow

- [X] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-279

### Reviewing help

In wagtail you should see a new section titled "Slice" on the domestic homepage as shown below:

Screenshot 2023-03-20 at 15 04 46

Populate this with a title and add three columns and fill in the fields. You should now see a new three card section on the homepage just below the hero banner as shown below:

Screenshot 2023-03-20 at 15 05 44

- [X ] Includes screenshot(s) - ideally before and after, but at least after

### Housekeeping

### Merging


core/templates/components/great/three_card_section.html
```diff
@@ -0,0 +1,17 @@
+<section class="great great-bg-white govuk-!-padding-top-5 govuk-!-padding-bottom-3">
+    <div class="govuk-width-container great-container">
+      {% if title %}
+        <h2 class="govuk-heading-l govuk-!-margin-bottom-5">{{ title }}</h2>
+      {% endif %}
+      <div class="govuk-grid-row">
+        {% if columns %}
```
mforner13, line 7: Not a biggie, but I think it might be overkill to wrap the card section in an if statement, as I think it's safe to assume this component would only be used if there were cards

domestic/migrations/0042_merge_20230414_1323.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 3.2.18 on 2023-04-14 13:23
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('domestic', '0040_alter_greatdomestichomepage_slice_columns'),
```
mforner13, line 8: My preference here would be to, after rebasing, delete the 3 migration files on this branch and create new ones so that we don't have files with the same number/separate branches of dependencies - what do you think? :)

styles/great/_components.scss
```diff
@@ -118,3 +118,36 @@
...
+        display: block;
+        font-size: 16px;
+
+        @include govuk-media-query($from: tablet) {
+            min-height: 22px;
+        }
+    }
+}
+
+.great-hr {
```
mforner13, line 147: What about `great-hr-blue`? As it's likely we'll have different coloured hrs (I think we already do but just outside the `great/` directory)

tests/unit/domestic/test_models.py
```diff
@@ -1973,6 +1973,70 @@ def test_great_domestic_homepage_magna_ctas_labels(root_page, client, user):
...
+def test_great_domestic_homepage_slice(root_page, client):
+    homepage = GreatDomesticHomePageFactory(
+        parent=root_page,
+        slug='root',
+    )
+
+    SiteFactory(
+        root_page=homepage,
+        hostname=client._base_environ()['SERVER_NAME'],
+    )
```
mforner13, line 1986: We've got existing fixtures for the homepage and domestic site which you could use here :)
https://github.com/uktrade/great-cms/blob/develop/tests/conftest.py#L210

mforner13 (commented):
Looks good - a couple of suggestions

tests/unit/domestic/test_models.py
mforner13: https://github.com/uktrade/great-cms/blob/develop/tests/unit/learn/test_models.py#L14 :)

mforner13 (approved):
👍

PR #2065, opened 2023-03-27
Fix show more component to work cross-browser

mforner13:
This PR updates the Show more component to work across current and older versions of chrome, firefox and edge. And also removes the additional arrow added next to the 'Show more' text in Safari.

To do this, it no longer relies on the css `has()` function which is not fully supported by these broswers/versions.

This change uses javascript to hide and reveal the additional event description text when the 'Show more' link is clicked, and displays the whole paragraph text when javascript is disabled.

**Javascript on - firefox**
Screenshot 2023-03-27 at 10 43 37

**Javascript off**
Screenshot 2023-03-27 at 10 44 24

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-497

### Reviewing help

### Housekeeping

### Merging


export_academy/templates/export_academy/event_list.html
```diff
@@ -79,6 +82,7 @@ <h4>{{ event.name }}</h4>
 {% block body_js %}
 {{ block.super }}
 <script src="{% static 'javascript/eventFilters.js' %}"></script>
+<script src="{% static 'javascript/showMore.js'%}"></script>
```
mforner13, line 85: Any suggestions on a better way to link the show more js to the html component so that they are always used together are welcome!

core/templates/components/show_more.html
```diff
@@ -1,5 +1,8 @@
+{% comment %} Uses domestic/static/javascript/showMore.js
+to reveal full text when user clicks summary link {% endcomment %}
+
 <div class="show-more">
-    <p id="details-text">{{ text }}</p>
+    <p class='details-text details-text-closed'>{{ event.description }}</p>
     <details closed class="govuk-details">
```
lewis-coulson-dit, line 6: i'd put a data-attr here

domestic/static/javascript/showMore.js
```diff
@@ -0,0 +1,9 @@
+const handleToggle = (element) => {
+  const text = element.previousElementSibling
+  return (
+    text.classList.toggle("details-text-open")
+  )
+}
+
+const detailsElements = document.querySelectorAll('details')
```
lewis-coulson-dit, line 8: i'd say use a data-attr just in case in future if we use the details element and don't want this behaviour / want to use the standard details implementation with text inside the details tag

lewis-coulson-dit (changes requested):

lewis-coulson-dit (approved):
LGTM

PR #2123, opened 2023-04-25
Download ics file for export academy events

bsabhilash:


lewis-coulson-dit (approved):
LGTM

export_academy/views.py
```diff
@@ -103,3 +110,39 @@ def get_context_data(self, **kwargs):
...
+        meeting['UID'] = uuid4()
+
+        description = f'{event.name}\n\n{event.description}{calender_content()}'
+        meeting.add('DESCRIPTION', description)
+
+        file_name = get_valid_filename(event.name)
+        alarm = Alarm()
+        alarm.add('action', 'DISPLAY')
+
+        alert_time = timedelta(minutes=-15)
```
timothyPatterson, line 140: Should this be something like:
``alert_time = event_time - timedelta(minutes=15)``\
Just a question - difficult to tell from docs.

bsabhilash: No. As per the RFC for iCalendar, we need to give relative time, and client will pick the event start time from event

timothyPatterson: Perfect thanks

PR #2125, opened 2023-04-25
Hotfix backport to CSS bullet points missing

rafa-garcia:
### Merging

### DIFF
--- a/domestic/templates/domestic/includes/article_detail_content.html
+++ b/domestic/templates/domestic/includes/article_detail_content.html
@@ -69,7 +69,7 @@
         {% endif %}
 
         <section class="grid-row padding-bottom-60">
-          <div class="article-content">
+          <div class="article-content rich-text">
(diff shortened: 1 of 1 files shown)


domestic/templates/domestic/includes/article_detail_content.html
```diff
@@ -69,7 +69,7 @@ <h2 class="heading-small">Tags</h2>
         {% endif %}
 
         <section class="grid-row padding-bottom-60">
-          <div class="article-content">
+          <div class="article-content rich-text">
```
LuptonM, line 72: i am not sure how it might break the form inside the stream field

Instead i would go into the streamfield html

and add the block-text class around the block so the form is unaffected

<div class="block-text">

        {{ block }}

    </div>

rafa-garcia: This is only a backport of a hotfix. If a refactoring needs to be done then a new ticked has to be opened.

PR #2170, opened 2023-05-17
EAM-230: tweaks to card after design review

lewis-coulson-dit:
Minor layout tweaks to EA card on logged in dashboard page

### Merging


styles/great/_utils.scss
```diff
@@ -48,6 +48,11 @@
         color: $great-dark-grey;
     }
 
+    .govuk-body {
```
mforner13, line 51: out of interest - why `govuk-body` here and not `great-body`?

lewis-coulson-dit: so we get all the responsive css plus all other sensible defaults in `govuk-body` but just with the tweak of settigng font size to 20px which is out standard size for body copy currently

mforner13 (approved):

PR #2227, opened 2023-06-08
Fix template auto-reloading

mforner13:
This PR fixes an issue which, since upgrading to django 4.1.9, meant that we had lost auto-reloading after changing an html template. The issue was caused by the `runserver_plus` command coming from `django-extensions` package, and fixed in version 3.2.3.

The newer version now calls the `django.template.autoreload.get_template_directories` method, which in turn calls [django.utils._os.to_path](https://github.com/django/django/blob/main/django/utils/_os.py#L56). This method relies on paths being either in string format or an instance of `pathlib.Path`. Because we create our `ROOT_DIR` path in settings.py using `environ.path` - this was causing `get_template_directories()` to give an error.

So this PR also updates our directory paths in settings.py to use `pathlib.Path`. This now looks to be the standard recommended in the [django docs here](https://docs.djangoproject.com/en/4.1/howto/overriding-templates/#overriding-from-the-project-s-templates-directory).

To test
- Install requirements - `make manage install_requirements`
- Run `make webserver`
- Go to any page in your browser
- Update the template for that page
- You should immediately see changes in your browser after refreshing the page

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-757

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -1,5 +1,6 @@
 import os
 import sys
+from pathlib import Path
 
 import directory_healthcheck.backends
 import environ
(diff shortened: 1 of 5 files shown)


ecbm (approved):
lgtm! 👍

PR #2233, opened 2023-06-13
DPE-162: third step

lewis-coulson-dit:
Adds the third form the new "Digital point of Entry" contact support form.

Screenshot 2023-06-13 at 10 23 16

To test locally go to "http://greatcms.trade.great:8020/contact/domestic/export-support/"

Fill in the first and second step of form and then click "Continue" button. You will be redirected to the third step.

On successful submit the form redirects to an empty page where the fourth step form will be located in a future ticket.

### Workflow

- [X] A clear/description pull request messaged added.

### Merging


contact/templates/domestic/contact/includes/govuk-form-field.html
```diff
@@ -1,5 +1,7 @@
 {% if field %}
-<label class="govuk-label govuk-!-margin-bottom-2 great-font-bold" for="{{field.id_for_label}}">{{field.label}} {% if optional %}<span>(Optional)</span>{% endif %}</label>
+<label class="govuk-label govuk-!-margin-bottom-2 great-font-bold" for="{{field.id_for_label}}">
+    {{field.label}} {% if optional %}<span>(Optional)</span>{% endif %}{% if dynamicText %}at [{{dynamicText}}]{% endif %}
```
mforner13, line 3: Having the `at` before the dynamic text seems quite specific to the job title field. Could you include at in the dynamic text you pass through from the template? So `at <job_title>`

contact/views.py
```diff
@@ -281,18 +284,31 @@ def form_valid(self, form):
...
+        if self.request.session.get('form_data'):
+            form_data = pickle.loads(bytes.fromhex(self.request.session.get('form_data')))[0]
+
         return super().get_context_data(
             **kwargs,
             heading_text='About you',
+            strapline_text='This information will allow us to contact you about your enquiry.',
             button_text='Continue',
             step_text='Step 3 of 7',
             show_form=settings.FEATURE_DIGITAL_POINT_OF_ENTRY,
```
mforner13, line 301: I know this isn't new to this PR - but would we want to hide the URLs behind the feature flag rather than the forms? If a user somehow (not sure how) got hold of the URL, it could be strange for them to just see the title of the page.
You can hide all of them at once by adding logic to `contact/urls.py` like in [this example](https://github.com/uktrade/great-cms/blob/develop/config/urls.py#L79).

lewis-coulson-dit: would I need to create a new app just for the digital point of entry urls? currently they live with the rest of "contact" urls

contact/templates/domestic/contact/includes/govuk-form-field.html
lewis-coulson-dit: ha yea that's what i wanted to do but I don't think i could pass a string with var in the include block like this `dynamicText='at [{{form_data.business_name}}]'`

contact/views.py
mforner13: No you can just add the if statement in the contact urls file :) something like:
```
if settings.FEATURE_DIGITAL_POINT_OF_ENTRY:
    urlpatterns += [
         path(
                 'contact/domestic/export-support/step4/',
                 skip_ga360(DomesticExportSupportFormStep4View.as_view()),
                 name='export-support-step-4',
             ),
             ...
    ]
```

contact/templates/domestic/contact/includes/govuk-form-field.html
lewis-coulson-dit: just found a solution on stack overflow

PR #2239, opened 2023-06-14
Update/kls 730 persist language across microsites

bobby-didcoding:
This PR is to enable language persistence across microsite subpages. I have created a new custom template tag that is used to append a language url param if one has been used in the parent page.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-730

### Merging


Dockerfile.dev
```diff
@@ -0,0 +1,35 @@
+FROM python:3.9.2
```
mforner13, line 1: Was this meant to be part of this PR? :) Could be good to make a separate one to update docker usage

domestic/templatetags/component_tags.py
```diff
@@ -269,3 +269,10 @@ def append_past_year_seperator(events):
             event.past_year_seperator = None
 
     return events
+
+
+@register.simple_tag
+def persist_language(url, language=None):
+    if language:
+        return f'{url}?lang={language}'
+    return url
```
mforner13, line 278: I think a [custom template filter](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#writing-custom-template-filters) could be more appropriate for this use case as they're used to modify a piece of data in place. 

It could then read like `<a href="{% link.value.full_url|persist_language :request.GET.lang %}"` which makes it a bit clearer what's going on :)

bobby-didcoding: Yeah, good point. I'll make the adjustment :)

PR #2303, opened 2023-06-29
Resolve EA user registration from unique link for sign up/in

mforner13:
This PR updates the sign in and sign up flows for users already registered to the Export Academy to fetch the user's registration object (and pre-populate their email address in the template) using a combination of `idb64` and `token` query parameters.

The idb64 value represents a user's aventri registration id, base 64 encoded.
The token represents a hashed version of the same user's email address.

To get the correct registration object:
- the aventri id (ie `Registration.external_id`) is decoded
- the user with that external_id is found
- the email address related to that registration object is hashed
- this hashed email is compared to the token in the query params
- if these values match, the user's email address is pe-populated in the sign in/up form

**To test locally**
- make sure you have directory-forms-api, directory-sso and directory-sso-proxy running
- ensure you have a registration in your local db with an external_id
- In a django shell run the following code:
```
import base64
import hashlib

me = Registration.objects.get(email='your_email')
token = hashlib.sha256(me.email.encode('UTF-8')).hexdigest()
idb64 = base64.b64encode(bytes(me.external_id, 'utf-8')).decode("utf-8")
``` 
- To sign in, go to http://greatcms.trade.great:8020/export-academy/signin?idb64=idb64&token=token, replacing the idb64 and token values with the ones you've just created in the shell
- To sign up, go to http://greatcms.trade.great:8020/export-academy/signup?idb64=idb64&token=token. You'll likely get redirected to the verification code page and get an email that says you're already registered 

**Sign in**
Screenshot 2023-06-29 at 16 13 35

**Sign up**
Screenshot 2023-06-29 at 16 15 18

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-802

### Reviewing help

### Merging

### DIFF
--- a/export_academy/helpers.py
+++ b/export_academy/helpers.py
@@ -1,4 +1,3 @@
-import base64
 import hashlib
 from functools import wraps
 
@@ -8,6 +7,8 @@ from django.contrib.auth.views import redirect_to_login
 from django.db.models import Q
 from django.shortcuts import redirect
 from django.urls import reverse_lazy
+from django.utils.encoding import force_str
+from django.utils.http import urlsafe_base64_decode
 
 from core.urls import SIGNUP_URL
 from export_academy.models import Event, Registration
@@ -192,7 +193,7 @@ def get_sectors_list(sector: str, second_sector: str, third_sector: str) -> str:
 
 
 def get_registration_from_unique_link(idb64, token):
-    external_id = base64.b64decode(idb64).decode('utf-8')
+    external_id = force_str(urlsafe_base64_decode(idb64))
     try:
         registration = Registration.objects.get(external_id=external_id)
         email_hash = hashlib.sha256(registration.email.encode('UTF-8'))
@@ -200,5 +201,5 @@ def get_registration_from_unique_link(idb64, token):
             return registration
         else:
             return None
-    except Exception:
+    except Registration.DoesNotExist:
         return None
--- a/export_academy/helpers.py
+++ b/export_academy/helpers.py
@@ -1,3 +1,5 @@
+import base64
+import hashlib
 from functools import wraps
 
 from django.conf import settings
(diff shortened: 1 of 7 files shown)


export_academy/helpers.py
```diff
@@ -178,3 +180,16 @@ def calender_content():
         'E: exportacademy@trade.gov.uk <mailto:exportacademy@trade.gov.uk>\n'
         'T: +44 (0) 2045 665 651'
     )
+
+
+def get_registration_from_unique_link(idb64, token):
+    external_id = base64.b64decode(idb64).decode('utf-8')
```
rafa-garcia, line 186: Could be better to URL-safe decode `idb64` like we do [here](https://github.com/uktrade/directory-sso/blob/db169ec393c3521a46f688d501b161406c74dc32/sso/verification/views.py#L47)

export_academy/helpers.py
```diff
@@ -178,3 +180,16 @@ def calender_content():
...
+    try:
+        registration = Registration.objects.get(external_id=external_id)
+        email_hash = hashlib.sha256(registration.email.encode('UTF-8'))
+        if email_hash.hexdigest() == token:
+            return registration
+        else:
+            return None
+    except Exception:
```
rafa-garcia, line 194: If catching any exception is the real intent, that's fine. However, unforeseeable things might happen throughout lines 188-190.

Also, attack attempts using query parameters would be hard to tell if we don't catch foreseeable exceptions only.

export_academy/mixins.py
```diff
@@ -88,31 +88,26 @@ def get_resend_verification_link(self):
...
 
 class HandleNewAndExistingUsersMixin:
-    def user_ea_registered(self):
-        # TODO update to handle unique token
-        registration_id = self.request.GET.get('registration-id')
-        if registration_id:
-            try:
-                if Registration.objects.get(pk=registration_id):
-                    return True
-            except Exception:
```
davidu1975, line 98: maybe catch Registration.DoesNotExist here

export_academy/helpers.py
mforner13: Very good point! I was thinking that external_id was a UUID field, so it would also catch the error if it was an invalid uuid, but realised it's just a charfield - so will update to be `Registration.DoesNotExist`

rafa-garcia (approved):
Nice one 👍🏻

PR #2320, opened 2023-07-05
Redirect middleware for handling Querystring

bsabhilash:
This PR overrides default RedirectMiddleware with custom one for handling querystring in redirects


config/settings.py
```diff
@@ -102,7 +102,7 @@
     'django.middleware.security.SecurityMiddleware',
     'whitenoise.middleware.WhiteNoiseMiddleware',
     'wagtail.contrib.legacy.sitemiddleware.SiteMiddleware',
-    'wagtail.contrib.redirects.middleware.RedirectMiddleware',
+    # 'wagtail.contrib.redirects.middleware.RedirectMiddleware',
```
bsabhilash, line 105: This redirect middleware is disabled and code added to `UserSpecificRedirectMiddleware`

hal274: @bsabhilash commented code is dangerous code. Can you delete it instead of commenting? What's the impact of someone accidentally uncommenting this line?

hal274: Thanks @bsabhilash - a backlog item (future work) reference with the detail of what was done (PR link) and what needs to be done or confluence doc is less risky IMHO

bsabhilash: Comment removed

PR #2335, opened 2023-07-10
Fix notifications on Save action for events

bsabhilash:
Update EA events to send notifications only when event is completed


ecbm (approved):
lgtm! 👍

PR #2369, opened 2023-07-19
EA sign up button for logged out users

bsabhilash:
EA sign up button for logged out users
Screenshot 2023-07-19 at 15 03 32


ecbm (commented):
lgtm! 👍

export_academy/helpers.py
```diff
@@ -79,16 +79,19 @@ def user_booked_on_event(user, event):
...
-            ]
+    if event.status is not Event.STATUS_FINISHED and not event.completed and not event.closed:
+        button = {
+            'label': f'Book<span class="great-visually-hidden"> {event.name}</span>',
+            'classname': 'govuk-button govuk-!-margin-bottom-0 ukea-ga-tracking',
+            'value': 'Confirmed',
+            'type': 'submit',
+        }
+        if user.is_anonymous:
+            button['label'] = f'Sign up<span class="great-visually-hidden"> {event.name}</span>'
```
jrbarnes9, line 90: I'm not sure we'd want to append the 'sign up' button with the vh event name, because they're not signing up to the event, rather than signing up to great/UKEA (to allow them to book an event). Might need new content from Neil.

bsabhilash: In my understanding, that event name is there for tracking who is clicking which event to get the dropout rate for the signup journey.

ecbm (approved):
lgtm! 👍

PR #2536, opened 2023-09-07
Event details page

LuptonM:
Testing (already gone through multiple pm reviews):

https://great.staging.uktrade.digital/export-academy/events/

Ticket

https://uktrade.atlassian.net/jira/software/projects/KLS/boards/359?assignee=63343adf88ed2ebef97c64ac&selectedIssue=KLS-1044

### Workflow

- [ ] A clear/description pull request messaged added.

### Reviewing help

### Housekeeping

- [ ] I have checked that my PR is using the latest package versions of: great-components, directory-constants, directory-healthcheck, directory-validators, directory-components, directory-api-client, directory-ch-client, django-staff-sso-client, directory-forms-api-client, directory-sso-api-client, sigauth

### Merging


export_academy/helpers.py
```diff
@@ -57,19 +59,45 @@ def update_booked_user_buttons(event, result, on_confirmation):
...
+                'label': 'Ended',
+                'classname': 'great-badge ended govuk-!-margin-bottom-0',
+            }
+        ]
+
+    elif (
+        is_export_academy_registered(user)
+        and not event_has_ended(event)
+        and not event.completed
+        and user_booked_on_event(user, event)
```
mforner13, line 78: (For after release) Would be good to extract this logic into a function or property as well to make more readable.

export_academy/helpers.py
```diff
@@ -57,19 +59,45 @@ def update_booked_user_buttons(event, result, on_confirmation):
...
+                'label': 'Closed',
+                'classname': 'great-badge closed govuk-!-margin-bottom-0',
+            }
+        ]
 
     return result
 
 
 def user_booked_on_event(user, event):
+    if user == AnonymousUser():
```
mforner13, line 99: I think you can just do `if user.is_anonymous` here

export_academy/views.py
```diff
@@ -606,6 +609,88 @@ def get_context_data(self, **kwargs):
         return ctx
 
 
+class EventsDetailsView(DetailView):
+    template_name = 'export_academy/event_details.html'
+    model = models.Event
+
+    def get_event_types(self, event):
```
mforner13, line 616: (For after release) I think this should also be a property on the event model to make this more widely usable. So you could call `event.type_names` for example.

export_academy/views.py
```diff
@@ -606,6 +609,88 @@ def get_context_data(self, **kwargs):
...
+        return [item.name for item in event.types.all()]
+
+    def get_warning_text(self):
+        def get_video_text(self):
```
mforner13, line 620: This method should live at the same level as `get_warning_text`

export_academy/views.py
```diff
@@ -606,6 +609,88 @@ def get_context_data(self, **kwargs):
...
+                    return ' Event recordings are only available for 4 weeks after the event.' if self.booked else ''
+                else:
+                    return ' Event recordings are only available for attendees to view for 4 weeks after the event.'
+            return ''
+
+        if self.ended or self.event.completed:
+            return 'This event has ended.' + get_video_text(self)
+
+        elif self.booked:
+            return ''
```
mforner13, line 632: Since you're already returning an empty string at the end, you shouldn't need this extra elif, as the scenaio of `self.booked` is being handled by that final return.

export_academy/views.py
```diff
@@ -606,6 +609,88 @@ def get_context_data(self, **kwargs):
...
+        ctx['has_event_badges'] = len(self.get_badges_for_event(self.event)) > 0
+        return ctx
+
+    def get_buttons_for_event(self, event):
+        user = self.request.user
+        return get_buttons_for_event(user, event)
+
+    def get_badges_for_event(self, event):
+        user = self.request.user
+        return get_badges_for_event(user, event)
```
mforner13, line 668: Is this method used anywhere?

tests/unit/export_academy/test_helpers.py
```diff
@@ -108,14 +122,33 @@ def test_join_button_returned_for_booked_in_progress_event(user):
...
+            'label': """Join event<span class="great-visually-hidden">opens in new tab</span>
+            <i class="fa fa-external-link-alt govuk-!-margin-right-0 govuk-!-margin-left-2" aria-hidden="true"></i>""",
             'title': 'Join Test event name',
             'url': reverse_lazy('export_academy:join', kwargs=dict(event_id=event.id)),
         }
     ]
 
 
+@pytest.mark.django_db
+def test_join_button_returned_for_booked_in_progress_event_but_no_link(user):
```
mforner13, line 135: The name of the test implies that the join button would be returned from `get_buttons_for_event` but then it asserts that an empty list is returned. Is this right? If so, is worth updating the name of the test.

mforner13 (approved):
I've left some comments to clean some things up, but they shouldn't block the release so could addressed in an additional PR after :)

export_academy/templates/export_academy/includes/event_warning.html
```diff
@@ -0,0 +1,11 @@
+<div class="event-warning-wrapper">
```
mforner13, line 1: Since this is only used in one place, i don't think it's worth having in a separate includes. I think you could just have it in `event_details_banner.html`

PR #2544, opened 2023-09-08
Slug for detail page

bsabhilash:


LuptonM (commented):
in event_video html and in events_details html there are hrefs with event.id

can you update these files as well?

LuptonM (approved):

PR #2631, opened 2023-10-05
fix to hide H! but allow screen readers to read

bobby-didcoding:
PR to fix a H! element that needs to be hidden but visible to screen readers.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-1242

### Housekeeping

### Merging

- [x] This PR can be merged by reviewers


core/components/sass/components/hero/_base.scss
```diff
@@ -114,9 +114,9 @@
   image-rendering: auto;
 
   .great-hero-text {
-    h1 {
-      visibility: hidden
-    }
+    // h1 {
+    //   visibility: hidden
+    // }
```
sebs-code, line 119: Should we just delete this, rather than comment it out?

react-components/dist/common.js.map
sebs-code, line 1: Can I check.. when you generated these, did you pull the latest version of the Develop branch in and rebase? If not you might over write Marc and I's CSS changes that we merged into Develop this morning.

sebs-code: Oh wait.. when was this branch created? If it was after my merge earlier today, don't worry, should have it in.

bobby-didcoding: Good point....Leave it with me & I'll double check

core/components/sass/components/hero/_base.scss
bobby-didcoding: Thats an error on my part! It's missing some css so I'll adjust and push

PR #2729, opened 2023-11-15
Feature/kls 1264 vod tracking

davidu1975:
Enable Video On Demand page tracking for logged in Users
Only capture data the first time a user views the VOD page for an event, video pairing
Create Activity Stream View to enable data to be captured in Activity Stream and the Data Workspace

### Workflow

- [ ] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-1264 and KLS-1520
- [ ] A clear/description pull request messaged added.

### Housekeeping

### Merging


export_academy/migrations/0032_videoondemandpagetracking.py
```diff
@@ -0,0 +1,83 @@
...
+                (
+                    'modified',
+                    django_extensions.db.fields.ModificationDateTimeField(
+                        auto_now=True, null=True, verbose_name='modified'
+                    ),
+                ),
+                ('id', models.UUIDField(default=uuid.uuid4, primary_key=True, serialize=False)),
+                ('user_email', models.EmailField(max_length=254)),
+                ('hashed_uuid', models.CharField(max_length=200)),
+                ('region', models.CharField(blank=True, max_length=10, null=True)),
```
timothyPatterson, line 35: Is 10 long enough?

export_academy/views.py
```diff
@@ -774,7 +785,62 @@ def _user_has_accepted_cookies(self):
...
+        return location.get('region', None)
+
+    def _get_company_details(self, user):
+        if not isinstance(user, BusinessSSOUser):
+            return self.NO_COMPANY_INFO
+
+        company = user.company
+        if not company:
+            return self.NO_COMPANY_INFO
+        return (company.name, company.postcode, '')
```
timothyPatterson, line 801: I think a comment here would be helpful explaining that we are returning '' as third element because we don't capture company telephone number anywhere. Either that or remove as it is redundant.

timothyPatterson (approved):

export_academy/migrations/0032_videoondemandpagetracking.py
davidu1975: made it 50 now, think 10 is okay but changed to be sure

export_academy/views.py
davidu1975: Added comment in case we capture this field in the future

PR #2868, opened 2024-01-30
KLS-1795 add inline feedback component

timothyPatterson:
This PR adds an _inline_ feedback component to every page on the great.gov.uk domestic site. By inline we mean that the user is not directed away from their current page; note however that this is unavoidable where JavaScript is not enabled. Responses are saved in directory-forms-api via a SaveOnlyInDatabase action for downstream processing.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/jira/software/projects/KLS/boards/359?selectedIssue=KLS-1795

### Reviewing help

There are five possible states of the feedback component:

1. Initial state enabling the user to select high-level 'Yes'/'No' values for page usefulness
Screenshot 2024-01-30 at 16 37 21
2. Where the page was useful there is opportunity to provide more detailed positive feedback
Screenshot 2024-01-30 at 16 38 33
3. Where the page was not useful there is opportunity to provide more detailed negative feedback
Screenshot 2024-01-30 at 16 38 13
4. Upon successful receipt of feedback a confirmation status is shown
Screenshot 2024-01-30 at 16 40 56
5. If an error has been encountered, for example due to loss of internet connectivity an error status is shown
Screenshot 2024-01-30 at 16 37 41

### Housekeeping

### Merging


core/templates/components/inline_feedback/negative_feedback.html
```diff
@@ -0,0 +1,78 @@
+<span>
```
lewis-coulson-dit, line 1: this should be a `div` rather than an inline `span`. You can put a `span` to a `div` but not the other way around. You can just add a utility class to set the `div` to display inline.

core/templates/components/inline_feedback/negative_feedback.html
```diff
@@ -0,0 +1,78 @@
...
+    <div class="inline-feedback__submission-confirmation">
+        <span role="img" class="fa fa-check-circle"></span>
+        <h5>Thanks for letting us know</h5>
```
lewis-coulson-dit, line 4: You might want to check this is correct. My suspicion is you will probably need a h3/h4 here instead.

core/templates/components/inline_feedback/was_page_useful.html
```diff
@@ -0,0 +1,116 @@
...
+          method: 'POST',
+          headers: {
+            'Content-Type': "application/json",
+            'X-CSRFToken': csrfToken
+          },
+          body: JSON.stringify(body)
+        })
+      }
+
+      document.getElementById('page-useful-form').addEventListener('submit', async function(event){
```
lewis-coulson-dit, line 62: don't think you need the `async` here

styles/great/_components.scss
```diff
@@ -1755,3 +1755,71 @@ nav.great-breadcrumbs {
     }
   }
 }
+
```
lewis-coulson-dit, line 1758: Just prefix these with `great-` and update references in html.

styles/great/_components.scss
```diff
@@ -2006,3 +2006,90 @@ nav.great-breadcrumbs {
...
+  }
+
+  [type=checkbox] + label::before {
+    border: 2px solid #222222;
+    border-radius: 4px;
+  }
+
+}
+
+.govuk-grid-column-two-thirds--no-padding {
```
lewis-coulson-dit, line 2093: you shouldn't need this class, just use the govuk one `govuk-!-padding-0`

PR #2872, opened 2024-02-01
KLS 1815 - Add user input for product and market journey

boxadesign:
A new user journey to allow users to input (and for us to capture) the product they are selling and the market they are selling to.

The main changes are as follows:

**Hero on homepage**

- New hero added to the homepage under the feature flag `FEATURE_PRODUCT_MARKET_HERO`. The new hero contains an input which is a simple free text field where the user can enter anything or nothing.

- If the user enters something and clicks continue they will be taken to a new page to enter the market they are selling to.

- The product entered on the homepage is appended as a query parameter which is used to send the correct data to the page.

- If the user enters nothing, they will still go to the market page but the product value will not be used in the page.

**Market page (view 1)**

- The input on this page uses a typeahead component which provides a list of available countries after the user enters 2 or more characters. 

- When the user selects a country and clicks continue this will be appended as another query string along with any provided product.

**Market page (view 2)**

- The page displays confirmation text of the selected product and market and a link (with content) to the market page for the selected country.

**Forms**

- Product and market submissions are stored in forms api. If the user is logged in, their user ID will be stored with the submissions.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-1815

### Merging


lewis-coulson-dit (approved):
LGTM

PR #2982, opened 2024-03-18
Improving mappings for trade associations

stuart-mindt:
Too many trade associations are being shown for tech and smart cities sector, removed one keyword - makes little difference as this sector is so broad. Added mappings for other sectors to ensure some are shown. This is a first pass as EYBs BA is looking to add more mappings later. This was mainly a bug fix to show results for Healthcare services.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IOO-1191

### Reviewing help

### Housekeeping

N/A

### Merging


international_online_offer/core/helpers.py
```diff
@@ -99,8 +98,18 @@ def get_trade_assoication_sectors_from_sector(sector):
...
+        directory_constants_sectors.EDUCATION_AND_TRAINING: ['Education', 'Training'],
+        sectors.ADVANCED_ENGINEERING: ['Engineering', 'Advanced Materials & Metals'],
+        sectors.AGRICULTURE_HORTICULTURE_FISHERIES_AND_PETS: [
+            'Agriculture',
+            'Agriculture, Horticulture and Fisheries',
+            'Food and Drink Agriculture',
+            'Agriculture & Manufacturing',
+            'Pets',
+        ],
+        directory_constants_sectors.RAILWAYS: ['Rail'],
```
timothyPatterson, line 112: How large will these mappings grow in the future and are they subject to change? If so, at some point we may wish to consider a django model for the sector mappings so they can be edited without requiring a code change.

international_online_offer/core/scorecard.py
```diff
@@ -70,7 +70,7 @@ def is_labour_workforce_hire(sector, hiring):
         {directory_constants_sectors.FINANCIAL_AND_PROFESSIONAL_SERVICES: 11},
         {directory_constants_sectors.CONSUMER_AND_RETAIL: 15},
         {sectors.CREATIVE_INDUSTRIES: 9},
-        {sectors.HEALHCARE_SERVICES: 10},
+        {sectors.HEALTHCARE_SERVICES: 10},
```
timothyPatterson, line 73: Kind of related to above comment about using django models - because this code is open source I wonder is there potential for people to be 'creative' with their answers to optimise the score.

timothyPatterson (approved):

international_online_offer/core/helpers.py
stuart-mindt: They're based on the sector mappings in data workspace. Unfortunately they do not map directly the our DBT sector list meaning we need this in-between mapping also. They'll get updated a few times until our BA is happy with them. From there they'll be updated yearly. Agree we could look at how to improve this down the line

international_online_offer/core/scorecard.py
stuart-mindt: I think you and I need to discuss this more. Remind me down the line. I'd like a better solution for this as we have EYB constant sector list, directory constant sector list and then DBT Data workspace sector list to contend with. Ideally there'd be a single source (DW's) that we can reference in code.

PR #3029, opened 2024-04-22
FIX - Add new cookie notice to learn base template

boxadesign:
Cookie modal was not showing on the learn pages. This was due to the base template not using the new cookie confirmation banner.

### DIFF
--- a/lighthouserc.js
+++ b/lighthouserc.js
@@ -22,7 +22,7 @@ module.exports = {
       assertions: {
         'categories:performance': [
           'error',
-          { minScore: 0.50, aggregationMethod: 'pessimistic' },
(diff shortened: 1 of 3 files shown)


lighthouserc.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
       assertions: {
         'categories:performance': [
           'error',
-          { minScore: 0.50, aggregationMethod: 'pessimistic' },
```
hal274, line 25: We need to review the config as you suggested before minScore gets to zero

hal274 (approved):

lighthouserc.js
boxadesign: Yes its on its way down but I think that now its actually fetching assets it just highlights the poor performance of the site.

PR #3048, opened 2024-05-02
Ioo 1336 scorecard sub sectors

stuart-mindt:
IST function have updated the scoring figures for the sectors. This change is to update those figures but also support sub sectors where possible and if the need is there. Expectation is that if a match is found on a sub sector rule then this should be prioritised over the parents figure.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IOO-1336

### Reviewing help

### Housekeeping

N/A

### Merging


international_online_offer/core/scorecard.py
```diff
@@ -31,55 +31,67 @@ def get_upper_value(value_in):
...
+        # Introduction of sub sectors added here for scoring.
+        # We do not have enumns / consts for these as there are many.
+        # (DBT Sub sectors in data workspace)
+        {'Energy : Civil nuclear': 759999},
+        {'Energy : Oil and gas': 2219999},
+        {'Consumer and retail : Books, printed media and stationery': 1799999},
+        {'Consumer and retail : Clothing, footwear and fashion': 499999},
+        {'Technology and smart cities : Communications': 1079999},
+        {'Technology and smart cities : Hardware': 2399999},
+        {'Technology and smart cities : Software': 503999},
```
timothyPatterson, line 60: I think we should be moving these to a Django model with values entered via django admin for two reasons:
1) Given this code is open source an actor could use this information to have their investment labelled as high-value and therefore try to get 1:1 support.
2) Every change/addition to this list requires a code change / merge / release.

international_online_offer/core/scorecard.py
```diff
@@ -5,8 +5,8 @@
 # Scoring system takes input from EYB triage and calculates whether a user / investor is low or high value.
 # ISD use this to contact high value users for help setting up in the UK
 # The numbers are given to us from ISD and tranferred into this scoring system.
-def score_is_high_value(sector, location, hiring, spend, spend_other=0):
-    # Requirement from senior stakeholders was that we only score based on three metrics:
+def score_is_high_value(sector, dbt_sub_sector, location, hiring, spend):
+    # Requirement from stakeholders was that we only score based on three metrics:
```
timothyPatterson, line 9: Maybe use a doc string for these multi-line comments explaining the purpose of a function?

international_online_offer/core/scorecard.py
stuart-mindt: 1. I agree regarding the gamification comment but see it as a very low risk. 
2. I want to remove the 'need' for a developer change when these figures update but it was agreed a while back with our previous BA that the process would remain locked down behind engineering and not exposed in wagtail or django admin. I think we're could do this now though and agree a new process flow / update the diagrams for who owns / updates / validates these values. For this iteration think its best to test this sub sector flow this way and backlog tech debt task to move towards django admin / wagtail controlled scoring figures. 
happy to be go either way at this point. What are your thoughts?

timothyPatterson: IMO at this stage in the project we have scope and resource to avoid the technical debt and do this via django admin and document accordingly.

timothyPatterson: Agreed with @stuart-mindt to merge this and create a Jira ticket to use django-admin which will be picked up as a priority.

timothyPatterson (approved):

stuart-mindt:
Going to improve the scoring in this backlog task: https://uktrade.atlassian.net/browse/IOO-1364

Move it into Django admin

PR #3052, opened 2024-05-03
Fix footer links for privacy and cookies in base footer templates for international homepage and campaigns

boxadesign:
Small fix to the international homepage and campaign site footer links


FoxJamie16 (approved):
LGTM

PR #3067, opened 2024-05-08
Notification Banner from design system 

divyaparameswaran:
Update the cookie notification banner to use the one provided by design system 

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GRTLV-37

### Reviewing help

Screenshot 2024-05-08 at 15 18 32

### Merging


boxadesign:
So I've added the .great class to the notification banner (on divya's branch) to bring in the styles that interact with govuk styles such as our font override etc.
 
Number of problems as seen below:
 
image

 
As a quick fix we will need to add some classes into our .utils.scss file to cater for the issues above (adding the :not vibe thats going on). Obvious thing is font colour, but responsive font sizing is off too.
 
This is how it looks in the govuk DS (smaller viewport):
 
image

Also need to adjust the mark up a bit.
 
This to me, shows the bigger issue of how we're doing the great styles. Ideally we shouldn't by default affect a govuk component when we bring it in and apply the .great style class other than font family. Generally, our styles aren't specific enough. An example is something like  .great h2 {} which will override any govuk style for a h2

boxadesign (approved):

PR #3104, opened 2024-05-21
Grtlv 33 bug on export academy form zendesk ticket 2

FoxJamie16:
- EAUK registration marketing choices form now uses radio select instead of dropdown. 
- Text input added to marketing options section of EAUK registration form, which only reveals when 'Other' is selected in the radio choice. 
- To accomplish above, conditional_reveal_id param added to radio-option.html to allow a radio select to reveal a div with aria-controls, this is used by the radios-with-other.html template from GreatRadioSelectWithOtherText widget

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GRTLV-33

### Reviewing help

### Housekeeping

- [x] I have checked that my PR is using the latest package versions of: great-components, directory-constants, directory-healthcheck, directory-validators, directory-components, directory-api-client, directory-ch-client, django-staff-sso-client, directory-forms-api-client, directory-sso-api-client, sigauth

### Merging

Screenshot 2024-05-21 at 16 38 46
Screenshot 2024-05-21 at 16 43 34


tests/unit/export_academy/test_forms.py
```diff
@@ -116,6 +116,10 @@
 )
 @pytest.mark.django_db
 def test_registration_form_validation(form, form_empty, error_messages):
+
```
FoxJamie16, line 119: These tests were slightly awkward to alter as it checks for an error message for any missing field, but the marketing_sources_other field is only required if marketing_sources=="Other", and only one error message should appear in any case. Solved by skipping when test case is not 'Other', though this also leaves success scenario untested in test_views.py. Suggestions on a better way to cover this would be appreciated.

divyaparameswaran (commented):
Hey Jamie , there might be some automated test fixes that might be needed .

export_academy/templates/export_academy/registration_base.html
```diff
@@ -50,7 +50,12 @@
         <div class="govuk-grid-row  govuk-!-padding-left-3 govuk-!-padding-right-3">
             <div class="great-grid-column-seven-twelfths">
                 <p class="govuk-body govuk-!-margin-top-4">{{ step_text }}</p>
-                <h1 class="govuk-heading-l govuk-!-margin-bottom-6">{{ title }}</h1>
+                {% if hide_title %}
```
lewis-coulson-dit, line 53: `{% if not hide_title %}`

bobby-didcoding (approved):
LGTM

PR #3269, opened 2024-08-01
breadcrumbs added to where to export and make an export plan

bobby-didcoding:
breadcrumbs added to core/base

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GRTLV-205

### Merging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,6 +148,17 @@ jobs:
           paths:
             - 'venv'
       - run:
+          name: Install Node.js
+          command: |
+            curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -
+            sudo apt-get install -y nodejs
+      - run:
+          name: Install @uktrade/great-design-system
+          command: |
+            npm install @uktrade/great-design-system@latest
+      - restore_cache:
+          key: v1-deps-{{ checksum "package-lock.json" }}
+      - run:
           name: Run Browser tests
           command: |
             python3 -m venv venv
--- a/tests/unit/core/test_wagtail_hooks.py
+++ b/tests/unit/core/test_wagtail_hooks.py
@@ -208,7 +208,7 @@ def test_estimated_read_time_calculation(rf, domestic_homepage):
     revision = detail_page.save_revision()
     revision.publish()
 
-    expected_duration = timedelta(seconds=181)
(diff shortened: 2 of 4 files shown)


tests/unit/core/test_wagtail_hooks.py
```diff
@@ -208,7 +208,7 @@ def test_estimated_read_time_calculation(rf, domestic_homepage):
     revision = detail_page.save_revision()
     revision.publish()
 
-    expected_duration = timedelta(seconds=181)
+    expected_duration = timedelta(seconds=182)
```
hal274, line 211: Just for my understanding, what is the difference between 181 and 182?

.circleci/config.yml
```diff
@@ -147,6 +147,17 @@ jobs:
           key: v1-deps-{{ checksum "requirements_test.txt" }}
           paths:
             - 'venv'
+      - run:
```
sebs-code, line 150: Should this be here for this PR? Looks more involved than breadcrumbs?

tests/unit/core/test_wagtail_hooks.py
sebs-code, line 211: Why change this?

bobby-didcoding: We have these estimated_read_duration methods peppered around the repo. they expect html to be read in xx seconds. Introducing/removing html such as breadcrumbs effect read time and  cause the test to fail.

.circleci/config.yml
bobby-didcoding: The change is needed to make breadcrumbs work. The repo is now looking for breadcrumb.html that is located in node_modules. The file does not exist without running npm install.

tests/unit/core/test_wagtail_hooks.py
bobby-didcoding: We have these estimated_read_duration methods peppered around the repo. they expect html to be read in xx seconds. Introducing/removing html such as breadcrumbs effect read time and cause the test to fail.

sebs-code: Gotcha - not criticising your increase (I can see why your had to do it), but the test itself seems absolutely insane to me. Will approve as it looks like the test has been here a long time.

sebs-code (approved):

PR #3361, opened 2024-09-02
Release PR

lewis-coulson-dit:
### Merging


core/management/commands/migrate_sector_tags.py
```diff
@@ -0,0 +1,66 @@
+from django.core.management import BaseCommand
+
+from core.models import SectorTag, SectorTagged
+from domestic.models import CountryGuidePage
+
+
+class Command(BaseCommand):
```
stuart-mindt, line 7: Directory API has an endpoint exposing the DBT sector list. I worry this is another example of where we're not aligning our datasets across DBT. Ive noticed that international also uses an older industry list which isnt the DW dbt sector list and we're looking to move away from this post PaaS migration. In a perfect world we should be using the DBT sector list in DW right across great. This is something the data science team are aligned with and have suggested we do on international. Another note is that data hub also uses the offical DBT sector list for CRM records.

stuart-mindt (approved):
Approved with notes for future work

PR #3380, opened 2024-09-09
Add hreflang and x-default links where appropriate

davidu1975:
Add hreflang and x-default links where appropriate

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GREATUK-1169

### Housekeeping

### Merging


contact/templatetags/contact_canonical_url_tags.py
```diff
@@ -9,3 +14,13 @@
 def get_canonical_url(context):
     request = context['request']
     return derive_canonical_url(request)
+
+
+@register.simple_tag(takes_context=True)
```
sebs-code, line 19: It looks like we have three very similar template tags in the PR, could we not just have one centrally located to avoid repetition, with some logic on the 'en' / 'en-gb' difference? Also would be good to have a comment string in these to explain the rationale behind them.

davidu1975: no does not work, cannot share tags between apps - tried

PR #3396, opened 2024-09-12
IGUK-380 - DnB integration (Backend)

timothyPatterson:
This PR introduces a 'lite' version of DataHub's Dunn & Bradstreet (D&B) proxy service with only required functionality migrated across and modified where appropriate. The change is necessary as there is a requirement to retrieve a company's D&B reference number to assist downstream consumers of EYB data. 

For context we were advised not to use DataHub's D&B proxy service because:
1. It does not support functionality such as typeahead search
2. There was a desire amongst stakeholders not to introduce dependancies between great-cms and DataHub
3. We only required a small subset of the functionality contained in the proxy service

For reference DataHub's D&B proxy service can be found [here](https://github.com/uktrade/dnb-service/tree/main).

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IGUK-380

### Housekeeping

### Merging


stuart-mindt (approved):
Looks good - if the api isnt available or returns nothing thinking the should app fail gracefully (when we do this work) Thinking that when we do the UI we could show / hide the field depending on this too.. Would be good that we aren't reliant on the D&B api for local dev.

PR #3484, opened 2024-10-09
Feature greatuk 1232 wagtail caching middleware

davidu1975:
## What
We are adding wagtail caching middleware and nopage_cache and cache_page decorators
## Why
To improve Great CMS performance response times

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GREATUK-1232

### Housekeeping

- [x] I have checked that my PR is using the latest package versions of: great-components, directory-constants, directory-healthcheck, directory-validators, directory-components, directory-api-client, directory-ch-client, django-staff-sso-client, directory-forms-api-client, directory-sso-api-client, sigauth

### Performance
- [x] Evaluated the performance impact of the changes
- [x] Ensured that changes do not negatively affect application scalability.

### Merging


config/settings.py
```diff
@@ -191,9 +193,18 @@
 else:
     REDIS_URL = env.str('REDIS_URL')
 
+# wagtail caching options
+# (see https://docs.coderedcorp.com/wagtail-cache/getting_started/django_settings.html#django-settings)
+WAGTAIL_CACHE = env.bool('WAGTAIL_CACHE', False)  # set to false for local
+WAGTAIL_CACHE_BACKEND = 'wagtail_cache'
```
hal274, line 199: Can this cache key be called ``great_wagtail_cache`` or something better that indicates great's non-default key?

config/settings.py
```diff
@@ -191,9 +193,18 @@
...
+WAGTAIL_CACHE_HEADER = True
```
hal274, line 200: Does ``WAGTAIL_CACHE_HEADER = True``this default to false? If no what does setting it explicitly gives us?

config/settings.py
davidu1975: sure

config/settings.py
davidu1975: "By default, an HTTP header named X-Wagtail-Cache is added to the response to indicate a cache hit or miss. To turn off this header, set WAGTAIL_CACHE_HEADER = False, or to customize the header set to a string."
yes True is the default - but like to explicitly define it - I can set it to False or remove it

hal274: Let's leave it. Frameworks can sometimes change default on new version that bites us

PR #3610, opened 2024-11-05
IGUK-381 return empty string if country not found

timothyPatterson:
## What
Small bugfix for [this PR](https://github.com/uktrade/great-cms/pull/3607). Related [Sentry error](https://sentry.ci.uktrade.digital/organizations/dit/issues/138874/?environment=dev&project=173&referrer=alert_email).
## Why
In this line: https://github.com/uktrade/great-cms/blob/133d0526300229d5284d5438fbd37d384810d711/international_online_offer/views.py#L310 we call the `get_country_display_name` method which in turn makes an api request to get the display name of a country given an ISO2 code. There is an edge case whereby a user signs up with a regular Great account and navigates to EYB. In this case they do not have a UserData object and so the api is called with an empty string which returns a 404. This bug fix handles this scenario gracefully.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IGUK-281

### Merging

### DIFF
--- a/international_online_offer/services.py
+++ b/international_online_offer/services.py
@@ -1,5 +1,7 @@
 from typing import Dict, List, Tuple
 
+from rest_framework import status
+
 from directory_api_client import api_client
 from international_online_offer.core import regions
 from international_online_offer.core.professions import (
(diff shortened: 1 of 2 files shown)


FoxJamie16 (approved):
LGTM

PR #3646, opened 2024-11-14
GRTLV-405 - Fix column layout on rtl for campaign sites

boxadesign:
## What

Small fix to add a grid class so layout is correct when using rtl language (campaign sites)


FoxJamie16 (approved):
LGTM

PR #3731, opened 2024-12-12
Bug greatuk 1459 change wagtail get doc setting

bobby-didcoding:
## What
Testing DBT platform has identified an issue with serving static docs. We currently use 'direct' (see explanation below). However, the S3 bucket in DBT platform is private which prevents direct access to files so we need to change to 'redirect
## Why
The WAGTAILDOCS_SERVE_METHOD setting Determines how document downloads will be linked to and served. Normally, requests for documents are sent through a Django view, to perform privacy checks and potentially other housekeeping tasks such as hit counting. To fully protect against users bypassing this check, it needs to happen in the same request where the document is served; however, this incurs a performance hit as the document then needs to be served by the Django server. In particular, this cancels out much of the benefit of hosting documents on external storage, such as S3 or a CDN.

For this reason, Wagtail provides several serving methods that trade some of the strictness of the permission check for performance:

'direct' - links to documents point directly to the URL provided by the underlying storage, bypassing the Django view that provides the permission check.

'redirect' - links to documents point to a Django view which will check the user’s permission; if successful, it will redirect to the URL provided by the underlying storage to allow the document to be downloaded. This is most suitable for remote storage backends such as S3, as it allows the document to be served independently of the Django server.

**Notes**
1) The new env variable defaults to 'redirect'. This is to avoid adding a new variable in DBT PaaS. 
2) I have added "WAGTAILDOCS_SERVE_METHOD": "direct" to dev, staging, uat and production in Vault.
3) This pr contains a few autoformat changes that have nothing todo with the above work

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GREATUK-1459

### Merging


davidu1975 (approved):
lgtm

PR #3805, opened 2025-02-04
IGUK-621 fix error summary link

timothyPatterson:
## What
Fix href on error link text.
## Why
Improvement highlighted as part of accessibility audit.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/IGUK-621

### Reviewing help

Screenshot 2025-02-04 at 12 03 02

### Merging

### DIFF
--- a/international_online_offer/templates/eyb/includes/error_summary.html
+++ b/international_online_offer/templates/eyb/includes/error_summary.html
@@ -7,9 +7,7 @@
                     {% for field in form %}
                         {% for error in field.errors %}
                             <li>
-                                {% if error_message_link %}
-                                    <a href="{{ error_message_link }}">{{ error }}</a>
-                                {% elif field.id_for_label %}
+                                {% if field.id_for_label %}
                                     <a href="#{{ field.id_for_label }}">{{ error }}</a>
                                 {% else %}
                                     <a href="#{{ field.auto_id }}">{{ error }}</a>
--- a/international_online_offer/templates/eyb/includes/error_summary.html
+++ b/international_online_offer/templates/eyb/includes/error_summary.html
@@ -7,7 +7,9 @@
                     {% for field in form %}
                         {% for error in field.errors %}
                             <li>
-                                {% if field.id_for_label %}
+                                {% if error_message_link %}
+                                    <a href="{{ error_message_link }}">{{ error }}</a>
+                                {% elif field.id_for_label %}
                                     <a href="#{{ field.id_for_label }}">{{ error }}</a>
                                 {% else %}
                                     <a href="#{{ field.auto_id }}">{{ error }}</a>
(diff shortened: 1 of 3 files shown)


international_online_offer/templates/eyb/includes/error_summary.html
```diff
@@ -7,7 +7,9 @@ <h2 class="govuk-error-summary__title">There is a problem</h2>
                     {% for field in form %}
                         {% for error in field.errors %}
                             <li>
-                                {% if field.id_for_label %}
+                                {% if error_message_link %}
+                                    <a href="{{ error_message_link }}">{{ error }}</a>
+                                {% elif field.id_for_label %}
```
peterhudec, line 12: The problem with this solution is that if there are multiple fields, possibly with multiple errors and if `error_message_link` is set, all of them will point to the same anchor.

The reason why the link is broken is that Django's `RadioSelect` assigns `id_for_label` suffixed with each radio's index when it iterates over its `choices`, but when accessed elsewhere e.g. in the error summary, it's without suffix.

I would prefer a more robust solution, like subclassing the `RadioSelect` with its `id_for_label` method overriden:
```py
class MyRadioSelect(RadioSelect):
    def id_for_label(self, id_, index=None):
        return f"{id_}_0" if index is None else super().id_for_label(id_, index)

class HiringForm(forms.Form):
    hiring = ChoiceField(
        label='How many people are you looking to hire in the UK?',
        required=True,
        # Having the id here didn't help as it also ended up suffixed
        # widget=RadioSelect(attrs={'id': 'hiring-select', 'class': 'govuk-radios__input'}),
        widget=MyRadioSelect(attrs={'class': 'govuk-radios__input'}),
        choices=choices.HIRING_CHOICES,
        error_messages={
            'required': 'Select how many people you want to hire',
        },
    )
```

peterhudec (commented):
It does the trick in this case, but it would break links to other fields if there were more of them.

international_online_offer/templates/eyb/includes/error_summary.html
timothyPatterson: Great comment - thanks. 

In this specific page we are following a one question per page pattern in the triage as per [GDS](https://design-system.service.gov.uk/patterns/question-pages/#start-by-asking-one-question-per-page) hence why the approach to error hrefs was taken. 

However, for some reason the implementation of this page differs from that elsewhere as throughout the triage we use `contact_widgets.GreatRadioSelect` which is a subclass of `django.forms.widgets.ChoiceWidget` that has the following `id_for_label` implementation which is similar to your proposal and works as expected:

```   
def id_for_label(self, id_, index="0"):
      """
      Use an incremented id for each option where the main widget
      references the zero index.
      """
      if id_ and self.add_id_index:
          id_ = "%s_%s" % (id_, index)
      return id_
```

So I have changed the implementation of this triage question to align with the others.

peterhudec (approved):
Looks good 👍

PR #3812, opened 2025-02-11
Release PR

timothyPatterson:


core/components/static/javascript/vendor/lodash.debounce.js
```diff
@@ -0,0 +1,375 @@
+/**
```
sebs-code, line 1: Why are we import this into the codebase rather than using a package manager?

core/geolocation_data/GeoLite2-Country.mmdb
sebs-code, line 1: We had some problems post migration with corrupt versions of this file causing failing tests. May be worth keeping an eye on.

core/static/javascript/accessible-autocomplete.v3.01.min.js
sebs-code, line 1: Same question - this looks like a third party library - why bring it into the repo vs. a package manager?

international_online_offer/management/commands/check_trade_association_links.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def handle(self, *args, **options):
+        for trade_association in TradeAssociation.objects.all():
+
+            try:
+                link_valid = self.check_page_found(trade_association.website_link)
+            except Exception:
+                link_valid = False
+
+            if trade_association.link_valid != link_valid:
```
sebs-code, line 43: Could possibly simplify this slightly:

        if trade_association.link_valid != link_valid:
            trade_association.update(link_valid=link_valid)

international_online_offer/templates/eyb/index.html
```diff
@@ -139,22 +139,19 @@ <h2 class="govuk-heading-l">
                     </h2>
                     <ul class="govuk-list">
                         <li class="govuk-grid-column-one-third govuk-!-margin-bottom-4">
-                            <div class="eyb-connection-image eyb-networking-image"
-                                 alt="Image of a person or potential connection"></div>
+                            <div class="eyb-connection-image eyb-networking-image"></div>
```
sebs-code, line 142: Does this need an alt tag? Same for two below.

sebs-code (approved):

core/components/static/javascript/vendor/lodash.debounce.js
timothyPatterson: This is a good point. We have followed this pattern in the past hence the vendor directory. I asked around when doing this work and it didn't seem possible to include an npm package into vanilla js code - but happy to be corrected and fix as tech debt.

core/static/javascript/accessible-autocomplete.v3.01.min.js
timothyPatterson: We adopted this pattern in the past and if you look in `core/static/javascript/` you will see two previous versions of this library. Obviously not ideal - probably a good tech debt candidate - but you can't remove previous versions without testing someone else's code.

international_online_offer/management/commands/check_trade_association_links.py
timothyPatterson: Good spot, thank you

international_online_offer/templates/eyb/index.html
timothyPatterson: @jrbarnes9 do you know?

PR #3815, opened 2025-02-12
Release PR

dbt-gs:


core/templates/captcha/includes/js_v3.html
```diff
@@ -1,4 +1,4 @@
-<script src="https://{{ recaptcha_domain }}/recaptcha/api.js?render={{ public_key }}{% if api_params %}&{{ api_params }}{% endif %}"></script>
+<script src="https://www.google.com/recaptcha/api.js?render={{ public_key }}"></script>
```
hal274, line 1: Any reason for hardcoding google domain here?

dbt-gs: The default value is `ww.google.com`. Unless google.com goes down this wouldn't work, so hardcoding it feels fine

dbt-gs: This code was not working before either, so the changes allows us to implement capture correctly

hal274: Agree with default being google.com but, it can be reconfigure using RECAPTCHA_DOMAIN to a different domain. I like the idea of updating env-var and redeploying a service over updating code. What is the difficulty or complexity of making this configurable instead of hardcoded?

PR #4064, opened 2025-04-30
BGST-409 privacy improvements

timothyPatterson:
## What
Increases privacy of User's triage data by:
1. Rename `session_id` to `triage_uuid` to better reflect the purpose of the column.
2. Use a new environment variable for the Fernet key. This is because setting the existing environment variable (`BETA_ENVIRONMENT`)  introduces middleware as a side-effect via [this conditional](https://github.com/uktrade/great-cms/blob/109b14475f1b555dd5ea84cd25d224022566cdca/config/settings.py#L914) which was not on any environment and fully investigating was outside the scope of this work.
3. When a user does not have cookies a query string parameter is passed between triage forms so we can identify the triage record to update. The previous approach used the raw triage_uuid as a query string parameter. This was determined to be a security risk as it could in theory be guessed and additionally where user's had accepted cookies and were returning via a Share Link it was a raw Django session key. This PR encrypts and decrypts the triage_uuid before placing in a query string parameter.
4. The identifier of a Share Link is now called `url_token` and is from `python.secrets.token_urlsafe`. This removes the need for encrypted triage_uuid's to be in the emailed Share Link.
5. Add a URL expiry time. This introduces a Time To Live on Share Links. Set at a high-default (180 days) - the actual value will be revisited after discussion with Stakeholders.
6. Small copy change on Starting a Business Location form
## Why
Addresses concerns raised in BGS Threat Model.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/BGST-409

### Performance
- [ ] Evaluated the performance impact of the changes
- [ ] Ensured that changes do not negatively affect application scalability.

Possible impact on performance due to encryption/decryption of query string parameters. Discussed with @rbtadbt .

### Merging


divyaparameswaran (approved):

domestic_growth/helpers.py
```diff
@@ -273,27 +281,57 @@ def get_change_sector_link(request: HttpRequest) -> str:
...
+        domestic_growth_models.StartingABusinessGuideEmailRecipient
+        if PRE_START_GUIDE_URL in request.path
+        else domestic_growth_models.ExistingBusinessGuideEmailRecipient
+    )
+    if request.GET.get('url_token', False):
+        try:
+            return email_recipient_model.objects.get(url_token=request.GET.get('url_token'))
+        except ObjectDoesNotExist:
+            return None
+    return None
```
hal274, line 323: Python func returns None by default so line 323 is superfulous

domestic_growth/helpers.py
```diff
@@ -258,8 +264,10 @@ def get_change_answers_link(request: HttpRequest) -> str:
     else:
         triage_start_url = reverse('domestic_growth:domestic-growth-existing-location')
 
-    if request.GET.get('session_id', False):
-        return triage_start_url + f"?session_id={request.GET.get('session_id')}"
+    if request.GET.get('triage_uuid', False):
+        return triage_start_url + f"?triage_uuid={request.GET.get('triage_uuid')}"
+    elif request.META.get('triage_uuid', False):
```
hal274, line 269: Not particular to this pr but all other our code: python's `dict.get` returns None if key is missing by default. This means `if request.META.get('triage_uuid')` is sufficient without setting default=False unless we are setting to a variable

hal274 (approved):
Some observations

domestic_growth/helpers.py
timothyPatterson: Thanks @hal274, I have addressed in [this PR](https://github.com/uktrade/great-cms/pull/4102) so marking as resolved.

domestic_growth/helpers.py
timothyPatterson: Thanks @hal274, I have addressed in [this PR](https://github.com/uktrade/great-cms/pull/4102) so marking as resolved.

